PR #2, opened 2020-10-15
Contracts: Improve modules backwards compatibility

facuspagnuolo:
The new proposed architecture is basically to have a cache for the protocol modules on each module, this could be seen as unnormalized information ofc, but it actually carries a few improvements:
- It allows us to support backwards compatibility
- It improves the gas usage for all the major functionalities of the protocol

To understand the thought process please read the docs addition


packages/evm/test/helpers/utils/errors.js
```diff
@@ -90,14 +90,19 @@ const CONTROLLER_ERRORS = {
...
+  MODULE_ALREADY_ENABLED: 'CTR_MODULE_ALREADY_ENABLED'
 }
 
 const CONTROLLED_ERRORS = {
   CONTROLLER_NOT_CONTRACT: 'CTD_CONTROLLER_NOT_CONTRACT',
   SENDER_NOT_CONTROLLER: 'CTD_SENDER_NOT_CONTROLLER',
-  SENDER_NOT_DISPUTES_MODULE: 'CTD_SENDER_NOT_DISPUTES_MODULE',
-  SENDER_NOT_CONFIG_GOVERNOR: 'CTD_SENDER_NOT_CONFIG_GOVERNOR'
+  SENDER_NOT_CONFIG_GOVERNOR: 'CTD_SENDER_NOT_CONFIG_GOVERNOR',
+  SENDER_NOT_ACTIVE_VOTING: 'CTD_SENDER_NOT_ACTIVE_VOTING',
```
sohkai, line 103: I couldn't find a test that used this error

docs/2-architecture/readme.md
```diff
@@ -55,3 +55,100 @@ The main entry point of the protocol is `AragonCourt`, this component inherits f
...
+Notes:
+- The Treasury module shouldn't reference any particular Dispute Manager instance, deploying a new Dispute Manager is not a problem
+
+
+### 2.5.4. Jurors Registry
+
+| Relies on module |           On           |    Expected behavior     |
+|------------------|------------------------|--------------------------|
+| DisputesManager  | `assignTokens`         | Any active Dispute Manager should be able to manage juror tokens |
+| DisputesManager  | `burnTokens`           | Any active Dispute Manager should be able to manage juror tokens |
```
sohkai, line 133: We can probably use the `"` format like the others

docs/2-architecture/readme.md
```diff
@@ -55,3 +55,100 @@ The main entry point of the protocol is `AragonCourt`, this component inherits f
 This allows to guarantee a single and immutable address to the users of the Court protocol. `AragonCourt` does not implement core logic, only the main entry points of the protocol where each request is forwarded to the corresponding modules of the `Controller` to be fulfilled.
 
 Detailed information about `AragonCourt` can be found in [section 4](../4-entry-points).
+
+## 2.5. Migration Strategies
```
sohkai, line 59: Perhaps we can go over these in a call to double check we understand the migration strategies available as well as the nuances in the notes (I'm having an especially hard time qualifying the language laid out in the notes at the moment).

packages/evm/contracts/court/controller/IModuleCache.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.5.8;
+
+
+interface IModuleCache {
+
```
sohkai, line 5: Minor nitpick: we could probably remove this newline?

packages/evm/contracts/court/controller/Modules.sol
```diff
@@ -0,0 +1,19 @@
...
+    bytes32 internal constant TREASURY = 0x06aa03964db1f7257357ef09714a5f0ca3633723df419e97015e0c7a3e83edb7;
+
+    // Voting module ID - keccak256(abi.encodePacked("VOTING"))
+    bytes32 internal constant VOTING = 0x7cbb12e82a6d63ff16fe43977f43e3e2b247ecd4e62c0e340da8800a48c67346;
+
+    // JurorsRegistry module ID - keccak256(abi.encodePacked("JURORS_REGISTRY"))
+    bytes32 internal constant JURORS_REGISTRY = 0x3b21d36b36308c830e6c4053fb40a3b6d79dde78947fbf6b0accd30720ab5370;
+
+    // Subscriptions module ID - keccak256(abi.encodePacked("SUBSCRIPTIONS"))
+    bytes32 internal constant SUBSCRIPTIONS = 0x2bfa3327fe52344390da94c32a346eeb1b65a8b583e4335a419b9471e88c1365;
```
sohkai, line 18: More of a side note, but I'm wondering if there was some reasoning behind this ordering to the constants (the docs are in a different order)?

packages/evm/contracts/disputes/DisputeManager.sol
```diff
@@ -44,6 +40,7 @@ contract DisputeManager is ControlledRecoverable, ICRVotingOwner, IDisputeManage
     string private constant ERROR_INVALID_ADJUDICATION_STATE = "DM_INVALID_ADJUDICATION_STATE";
     string private constant ERROR_ROUND_ALREADY_DRAFTED = "DM_ROUND_ALREADY_DRAFTED";
     string private constant ERROR_DRAFT_TERM_NOT_REACHED = "DM_DRAFT_TERM_NOT_REACHED";
+    string private constant ERROR_VOTER_WEIGHT_ZERO = "DM_VOTER_WEIGHT_ZERO";
```
sohkai, line 43: Haha, do you think this makes sense as a "rounds-related" error?

packages/evm/contracts/test/court/AragonCourtMock.sol
```diff
@@ -101,4 +102,24 @@ contract AragonCourtMock is AragonCourt, TimeHelpersMock {
         if (mockedTermRandomness != bytes32(0)) return mockedTermRandomness;
         return super._computeTermRandomness(_termId);
     }
+
+    function _setAndCacheModule(bytes32 _id, address _addr) private {
```
sohkai, line 106: Took me a bit to understand this, so might be worth leaving a note.

If I understand correctly, this actually "installs" this new module instance across all of the protocol's modules?

packages/evm/contracts/test/court/AragonCourtMock.sol
```diff
@@ -34,29 +34,30 @@ contract AragonCourtMock is AragonCourt, TimeHelpersMock {
     {}
 
     function setDisputeManager(address _addr) external {
-        _setModule(DISPUTE_MANAGER, _addr);
+        _setAndCacheModule(DISPUTE_MANAGER, _addr);
     }
 
     function setDisputeManagerMock(address _addr) external {
         // This function allows setting any address as the DisputeManager module
```
sohkai, line 41: Technically, this operation is only applying to the main Protocol contract, and not any of its modules, right?

packages/evm/contracts/court/controller/IModuleCache.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.5.8;
+
+
+interface IModuleCache {
```
sohkai, line 4: I'm a bit wary of using "cache" as the name, as I don't think it's accurate. This module is not really caching information—usually the term implies an ability to re-retrieve old information faster—but serving as a replicated list for currently installed protocol modules.

packages/evm/test/helpers/utils/modules.js
```diff
@@ -0,0 +1,27 @@
+const { ZERO_ADDRESS } = require('@aragon/contract-helpers-test')
+const { soliditySha3, padLeft, toHex, toChecksumAddress } = require('web3-utils')
+
+const MODULE_IDS = {
+  disputes: '0x14a6c70f0f6d449c014c7bbc9e68e31e79e8474fb03b7194df83109a2d888ae6',
```
sohkai, line 5: Wondering two points:

1. Do we have tests for these module IDs?
2. Can we build these with a keccak, so tests invariably blow up if we ever screw with these constants?

packages/evm/test/helpers/utils/modules.js
```diff
@@ -0,0 +1,27 @@
...
+  subscriptions: '0x2bfa3327fe52344390da94c32a346eeb1b65a8b583e4335a419b9471e88c1365'
+}
+
+const getCachedAddress = async (target, moduleId) => {
+  // The modules cache are stored at the 1st index of the controlled contract storage
+  const modulesCacheSlot = padLeft(1, 64)
+  // Parse module ID en hexadecimal and pad 64
+  const moduleIdHex = padLeft(toHex(moduleId), 64)
+  // The modules cache is a mapping indexed by modules ID
+  const moduleSlot = soliditySha3(moduleIdHex + modulesCacheSlot.slice(2))
```
sohkai, line 18: More of a note for later, but maybe this is a useful utility to extract out. I don't see us using it that often, but can be really useful when we do need to calculate a particular mapping's slot.

packages/evm/contracts/court/controller/Controlled.sol
```diff
@@ -12,15 +14,25 @@ import "../../disputes/IDisputeManager.sol";
 import "../../subscriptions/ISubscriptions.sol";
 
 
-contract Controlled is IsContract, ConfigConsumer {
+contract Controlled is IsContract, IModuleCache, Modules, ConfigConsumer {
```
sohkai, line 17: Nitpick, what do you think about always putting interfaces first?

packages/evm/contracts/court/controller/Modules.sol
```diff
@@ -0,0 +1,19 @@
+pragma solidity ^0.5.8;
+
+
+contract Modules {
```
sohkai, line 4: Maybe this would be better named as `ModuleIds`?

This is a slight aside, but I'm also wondering if we should name the concepts in the `controller` folder as "modules", e.g. `ModuleController`, etc? I'm not too worried, but it might help to demarcate these from the typical `OnlyOwner` / `Controller` authentication contracts.

packages/evm/contracts/court/controller/Controlled.sol
```diff
@@ -12,15 +14,25 @@ import "../../disputes/IDisputeManager.sol";
...
-    string private constant ERROR_SENDER_NOT_DISPUTES_MODULE = "CTD_SENDER_NOT_DISPUTES_MODULE";
+    string private constant ERROR_SENDER_NOT_ACTIVE_VOTING = "CTD_SENDER_NOT_ACTIVE_VOTING";
+    string private constant ERROR_SENDER_NOT_ACTIVE_DISPUTE_MANAGER = "CTD_SEND_NOT_ACTIVE_DISPUTE_MGR";
+    string private constant ERROR_SENDER_NOT_CURRENT_DISPUTE_MANAGER = "CTD_SEND_NOT_CURRENT_DISPUTE_MGR";
 
     // Address of the controller
     Controller internal controller;
 
+    // List of module caches indexed by ID
+    mapping (bytes32 => address) internal modulesCache;
```
sohkai, line 32: Could we make both the `controller` and this `modulesCache` public so they have getters?

It's not a big deal, especially if our contracts don't have room, but it may ease the burden on any possible debugging down the line.

packages/evm/contracts/court/controller/Controlled.sol
```diff
@@ -38,10 +50,28 @@ contract Controlled is IsContract, ConfigConsumer {
...
+        _;
+    }
+
+    /**
+    * @dev Ensure the msg.sender is the current DisputeManager module
+    */
+    modifier onlyCurrentDisputeManager() {
+        (address addr, bool disabled) = controller.getDisputeManager();
+        require(msg.sender == addr, ERROR_SENDER_NOT_CURRENT_DISPUTE_MANAGER);
+        require(!disabled, ERROR_SENDER_NOT_ACTIVE_DISPUTE_MANAGER);
```
sohkai, line 74: More of a side note, but going on the hypothetical here, is disabling the current dispute manager a valid state?

If so, is the main consequence to disallow new disputes, or would it cripple other parts of the system?

packages/evm/contracts/court/controller/Controlled.sol
```diff
@@ -88,58 +132,77 @@ contract Controlled is IsContract, ConfigConsumer {
...
+        return controller.getModulesGovernor();
+    }
+
+    /**
+    * @dev Internal function to fetch the address of the DisputeManager module
     * @return Address of the DisputeManager module
     */
     function _disputeManager() internal view returns (IDisputeManager) {
-        return IDisputeManager(controller.getDisputeManager());
+        return IDisputeManager(_getModuleCache(DISPUTE_MANAGER));
```
sohkai, line 154: Nitpick, I'm wondering if we should prefix these constants with `MODULE_ID_` or similar to avoid possible confusion?

packages/evm/contracts/court/controller/Controlled.sol
```diff
@@ -54,6 +84,20 @@ contract Controlled is IsContract, ConfigConsumer {
...
+    /**
+    * @notice Update the implementations cache of a list of modules
+    * @param _ids List of IDs of the modules to be updated
+    * @param _addresses List of module addresses to be updated
+    */
+    function cacheModules(bytes32[] calldata _ids, address[] calldata _addresses) external onlyController {
+        require(_ids.length == _addresses.length, ERROR_INVALID_MODULES_CACHE_INPUT);
+
+        for (uint256 i = 0; i < _ids.length; i++) {
+            modulesCache[_ids[i]] = _addresses[i];
```
sohkai, line 96: Going on the hypothetical again, do you think it'd be useful to include an `isContract()` check?

I'm not so worried about gas cost for updating modules, so this would rule out some potential mistakes but does disallow counterfactually-deployed modules (which I think are not a good idea anyway?).

packages/evm/contracts/court/controller/Controlled.sol
```diff
@@ -88,58 +132,77 @@ contract Controlled is IsContract, ConfigConsumer {
...
     function _configGovernor() internal view returns (address) {
         return controller.getConfigGovernor();
     }
 
     /**
-    * @dev Internal function to fetch the address of the DisputeManager module from the controller
+    * @dev Internal function to fetch the controller's modules governor
+    * @return Address of the controller's modules governor
+    */
+    function _modulesGovernor() internal view returns (address) {
```
sohkai, line 145: AFAIK this isn't being used anywhere yet?

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -38,13 +28,26 @@ contract Controller is IsContract, CourtClock, CourtConfig {
...
+
     // Governor addresses of the system
     Governor private governor;
 
-    // List of modules registered for the system indexed by ID
-    mapping (bytes32 => address) internal modules;
+    // List of current modules registered for the system indexed by ID
+    mapping (bytes32 => address) internal currentModules;
+
+    // List of all modules registered for the system indexed by address
```
sohkai, line 45:
```suggestion
    // List of all historical modules registered for the system indexed by address
```

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -38,13 +28,26 @@ contract Controller is IsContract, CourtClock, CourtConfig {
...
+    mapping (address => Module) internal allModules;
```
sohkai, line 46: Similar to `Controlled`, I'm wondering if it'd be worth making these public?

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -38,13 +28,26 @@ contract Controller is IsContract, CourtClock, CourtConfig {
...
+        bool disabled;      // Whether the module is disabled
+    }
+
     // Governor addresses of the system
     Governor private governor;
 
-    // List of modules registered for the system indexed by ID
-    mapping (bytes32 => address) internal modules;
+    // List of current modules registered for the system indexed by ID
+    mapping (bytes32 => address) internal currentModules;
```
sohkai, line 43: Similar to `Controlled`, I'm wondering if it's worth making this public?

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -253,6 +256,54 @@ contract Controller is IsContract, CourtClock, CourtConfig {
         }
     }
 
+    /**
+    * @notice Disabled module `_addr`
```
sohkai, line 260:
```suggestion
    * @notice Disable module `_addr`
```

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -253,6 +256,54 @@ contract Controller is IsContract, CourtClock, CourtConfig {
...
+        module.disabled = false;
+        emit ModuleEnabled(module.id, _addr);
+    }
+
+    /**
+    * @notice Sync modules' cache for a list of modules IDs based on their current address
+    * @param _modules List of modules addresses to be synced
+    * @param _ids List of IDs of the modules to be cached
+    */
+    function cacheModules(IModuleCache[] calldata _modules, bytes32[] calldata _ids) external onlyModulesGovernor {
```
sohkai, line 291: Just found it a bit weird that `_ids` wasn't first, as the `IModuleCache` interface's function has it first.

Other than that... I can see how `cacheModules` fits what this is doing, but I think we might be able to massage this a bit so it's nicer.

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -341,52 +392,81 @@ contract Controller is IsContract, CourtClock, CourtConfig {
...
     */
-    function getModule(bytes32 _id) external view returns (address) {
+    function getModule(bytes32 _id) external view returns (address addr, bool disabled) {
         return _getModule(_id);
     }
 
     /**
-    * @dev Tell the address of the DisputeManager module
-    * @return Address of the DisputeManager module
+    * @dev Tell the address of the current DisputeManager module
```
sohkai, line 428: Similar for the other natspec strings

```suggestion
    * @dev Tell the information for the current DisputeManager module
```

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -436,27 +519,37 @@ contract Controller is IsContract, CourtClock, CourtConfig {
...
     /**
-    * @dev Internal function to tell address of a module based on a given ID
-    * @param _id ID of the module being queried
-    * @return Address of the requested module
+    * @dev Internal function to tell the address of the current DisputeManager module
+    * @return Current address of the DisputeManager module
     */
-    function _getModule(bytes32 _id) internal view returns (address) {
-        return modules[_id];
+    function _getDisputeManager() internal view returns (address) {
```
sohkai, line 544: AFAIK these two internal functions aren't getting used anymore

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -436,27 +519,37 @@ contract Controller is IsContract, CourtClock, CourtConfig {
...
-    function _getDisputeManager() internal view returns (address) {
-        return _getModule(DISPUTE_MANAGER);
+    function _ensureModuleExists(Module storage _module) internal view {
+        require(_module.id != bytes32(0), ERROR_MODULE_NOT_SET);
     }
 
     /**
-    * @dev Internal function to tell the address of the Subscriptions module
-    * @return Address of the Subscriptions module
+    * @dev Internal function to tell the current address registered for a module based on a given ID
```
sohkai, line 530: Similar note as before as it's not just the current address anymore

sohkai (commented):
I have mostly small suggestions, otherwise, I'm mostly wondering if we can name the concept a bit better 😄.

If I understand correctly, the replicated module storage is necessary because it allows disabled modules to hold onto their old module references (on top of being a gas-optimization to avoid CALLs)?

It's hard for me to audit the exact authentication changes well at the moment, so you may have to walk me over it in a call!

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -243,7 +246,7 @@ contract Controller is IsContract, CourtClock, CourtConfig {
     /**
     * @notice Set many modules at once
     * @param _ids List of ids of each module to be set
-    * @param _addresses List of addressed of each the module to be set
+    * @param _addresses List of addresses of each the module to be set
     */
     function setModules(bytes32[] calldata _ids, address[] calldata _addresses) external onlyModulesGovernor {
```
sohkai, line 251: ⚠️ As discussed, we may want to add two parameters to allow setting the cache on the new module and updating the caches on other modules (the up and downs on the "module dependency tree").

packages/evm/contracts/court/controller/Controlled.sol
```diff
@@ -88,58 +132,77 @@ contract Controlled is IsContract, ConfigConsumer {
...
     }
 
     /**
-    * @dev Internal function to fetch the address of the Voting module owner from the controller
+    * @dev Internal function to fetch the address of the Voting module owner
     * @return Address of the Voting module owner
     */
     function _votingOwner() internal view returns (ICRVotingOwner) {
-        return ICRVotingOwner(address(_disputeManager()));
+        return ICRVotingOwner(_getModuleCache(DISPUTE_MANAGER));
```
sohkai, line 178: I didn't catch this earlier, but I'm wondering if the concept of a "Voting owner" will be confusing given that we're allowed to switch modules around.

Would it be better to make this assumption only in the Voting module? (E.g. by using `_disputeManager()` and casting it to `ICRVotingOwner`)

docs/2-architecture/readme.md
```diff
@@ -55,3 +55,100 @@ The main entry point of the protocol is `AragonCourt`, this component inherits f
...
+| Relies on module |           On           |    Expected behavior     |
+|------------------|------------------------|--------------------------|
+| JurorsRegistry   | `getJuror`             | Each period should work always with the same Jurors Registry instance |
+| JurorsRegistry   | `getJurorShare`        | " |
+| JurorsRegistry   | `claimFees`            | " |
+| JurorsRegistry   | `ensurePeriodBalance`  | " |
+
+Notes:
+- The Subscriptions module could checkpoint the Jurors Registry instance per period
+- Otherwise, it can trust that the Jurors Registry always reflects the same status among all its versions
```
sohkai, line 154: Ah, we can leave this as an issue, but it'd also be worth also adding the new events / external interfaces to the docs (perhaps by adding a new page for `Controlled`?).

sohkai:
In thinking more about this concept, what do you think about renaming it something like "linked modules" or "module linker"? Rather than just "cache" :).

packages/evm/contracts/disputes/DisputeManager.sol
facuspagnuolo: in a sense yes, we throw it when the given voter has not been picked for the round basically

packages/evm/contracts/court/controller/Modules.sol
facuspagnuolo: Addressed in 2c58136e3a92fb804203e746dd2af8eb9b5c659a

docs/2-architecture/readme.md
facuspagnuolo: I think we already covered this one, let me know otherwise and we can discuss what's missing

packages/evm/contracts/court/controller/IModuleCache.sol
facuspagnuolo: Addressed in bd44557836f61d1a2807db77e1577e1c27b664f4

docs/2-architecture/readme.md
facuspagnuolo: Addressed in 667aa7eb7781da79614a2e6730a1fd4dfded5c5a

packages/evm/test/helpers/utils/modules.js
facuspagnuolo: Yup, I actually did it [here](https://github.com/aragon/protocol/commit/d872e263b80e3160eb979a7a9c0db5602641511c#diff-18991661b2b688266b71ed7c8b025b372aee7adcc1614150530f9bb5f2195aa3L5-L9)

packages/evm/contracts/court/controller/Controlled.sol
facuspagnuolo: I guess I always do, here I think I got biassed cause `IsContract` starts with `I` haha
Addressed in 1be31e2ff36797fa5e296d4866b4f8f9eca6078f

packages/evm/contracts/court/controller/Modules.sol
facuspagnuolo: Addressed in 56d8d60ae5cbe039298a0918ebf99b18af9dc388 and in 3032e3dd9375db85c147b7daec8ef26cde5efd31 respectively

packages/evm/contracts/court/controller/Controlled.sol
facuspagnuolo: Addressed in eb89c29520855ff2dcf8c2fb76771eab4789865e and in b4dab8fbb12cce801fd977f3785da3ac7b4ad1c3 respectively

packages/evm/test/helpers/utils/modules.js
facuspagnuolo: Yup, however I ended up removing this in b4dab8fbb12cce801fd977f3785da3ac7b4ad1c3

packages/evm/contracts/court/controller/Controlled.sol
facuspagnuolo: Addressed in 9e4098b539d07f4ee0408c1d4327fdf6da26681f

packages/evm/contracts/court/controller/Controlled.sol
facuspagnuolo: > More of a side note, but going on the hypothetical here, is disabling the current dispute manager a valid state?

Yes, as a rule of thumb, I decided not to prevent anything from disabling/enabling modules, up to governor what to do, but everything is possible.

> is the main consequence to disallow new disputes, or would it cripple other parts of the system?

When you disable a module you're basically not allowing other modules to interact with it. If you disabled a dispute manager, even if it's the current one, you're blocking the dispute, guardians from voting, and their corresponding locked balances in the registry.

packages/evm/contracts/court/controller/Controlled.sol
facuspagnuolo: I avoided that intentionally cause this is actually triggered by the modules set in the controller. And these are already checked to be contracts when being set

packages/evm/contracts/court/controller/Controlled.sol
facuspagnuolo: Nope, let's remove it. Addressed in 15536accca98a419387e1b60af5e28f5d8cd152c

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: These can be fetched using `getModuleByAddress`

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: These can be fetched using `getModule`

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: Addressed in 02c5d46b1692d9b16fca293326d3af02ee7d5174

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: Addressed in 02c5d46b1692d9b16fca293326d3af02ee7d5174

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: Addressed in 02c5d46b1692d9b16fca293326d3af02ee7d5174

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: Addressed in 02c5d46b1692d9b16fca293326d3af02ee7d5174

packages/evm/contracts/court/controller/Controlled.sol
facuspagnuolo: I liked that! Addressed in 04fb4fd595490ab273c8391103e2746027a6a92a

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: These are actually being used in the entry point but I decided to inline them in d19b99d57d35661855a34a0ad9696716dc4171a1

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: Addressed in cb92409c82d788fd8d297ffa83546311ab061829

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: The thing is that from the controller's point of view it's like "update these modules (addresses)  based on the addresses stored for these other modules (ids)", while for the cache module it's sth like "cache this module ID with this address".

packages/evm/contracts/test/court/AragonCourtMock.sol
facuspagnuolo: Yes, the main reason is because this is being used to force a controlled address as the dispute manager.
However, hope it has been clarified in 797f34c148e1085dc338ff724c6eae72649f251e

packages/evm/contracts/test/court/AragonCourtMock.sol
facuspagnuolo: Clarified in 797f34c148e1085dc338ff724c6eae72649f251e

packages/evm/test/helpers/utils/errors.js
facuspagnuolo: Addressed in ecefcb71a37dcdd663a56d7d46087fd8f3e43b2f

packages/evm/contracts/court/controller/IModuleCache.sol
facuspagnuolo: Renamed in 46d7b072a782becb650345133fceadc771e45bab following your idea bellow about the "linker"

docs/2-architecture/readme.md
facuspagnuolo: Not sure what do you mean here

facuspagnuolo:
I also implemented what we talked about allowing to interact with previous `DisputeManager` versions in 8f6c97ac5ee6931a685988520d07d6e393c6c0a3

docs/2-architecture/readme.md
sohkai: Let's discuss in a separate issue! https://github.com/aragon/protocol/issues/22

PR #3, opened 2020-10-15
Arbitration: Simplify interfaces

facuspagnuolo:
Proposing two major changes to both Arbitrable and Arbitrator interfaces:
- Move evidence logging to the Arbitrator
- Replace `executeRuling` from Arbitrator in favor of `rule` to fetch the ruling for a dispute
- Change `rule` from `Arbitrable` so it is not intended to be called from the arbitrator only
- With all these changes we can get rid of ERC165 - `Arbitrable#rule` is totally optional


packages/evm/test/helpers/utils/errors.js
```diff
@@ -1,6 +1,6 @@
 const ARAGON_COURT_ERRORS = {
   SUBSCRIPTION_NOT_PAID: 'AC_SUBSCRIPTION_NOT_PAID',
-  SENDER_NOT_ARBITRABLE: 'AC_SENDER_NOT_ARBITRABLE'
+  ERROR_SENDER_NOT_DISPUTE_SUBJECT: 'AC_SENDER_NOT_DISPUTE_SUBJECT'
```
sohkai, line 3: Wasn't sure, but I saw a couple of these prefixed with `ERROR_` and thought they might not need them since the object is already suffixed with `_ERROR` 🤷

packages/evm/contracts/arbitration/IArbitrator.sol
```diff
@@ -1,9 +1,18 @@
 pragma solidity ^0.5.8;
 
+import "./IArbitrable.sol";
 import "../lib/os/ERC20.sol";
 
 
 interface IArbitrator {
+    /**
+    * @dev Emitted when new evidence is submitted for the IArbitrable instance's dispute
```
sohkai, line 9:
```suggestion
    * @dev Emitted when new evidence is submitted for a dispute
```

packages/evm/contracts/arbitration/IArbitrator.sol
```diff
@@ -1,9 +1,18 @@
...
+    * @param disputeId Identification number of the dispute receiving new evidence
+    * @param submitter Address of the account submitting the evidence
+    * @param evidence Data submitted for the evidence of the dispute
+    */
+    event EvidenceSubmitted(uint256 indexed disputeId, address indexed submitter, bytes evidence);
```
sohkai, line 14: Should we also emit the related IArbitrable? Or is that referenceable by the `disputeId`?

packages/evm/contracts/court/AragonCourt.sol
```diff
@@ -106,13 +114,14 @@ contract AragonCourt is Controller, IArbitrator {
...
+    * @param _disputeId Identification number of the dispute to be ruled
+    * @return subject Arbitrable instance associated to the dispute
+    * @return ruling Ruling number computed for the given dispute
     */
-    function executeRuling(uint256 _disputeId) external {
+    function rule(uint256 _disputeId) external returns (IArbitrable subject, uint256 ruling) {
         IDisputeManager disputeManager = IDisputeManager(_getDisputeManager());
-        (IArbitrable subject, uint8 ruling) = disputeManager.computeRuling(_disputeId);
-        subject.rule(_disputeId, uint256(ruling));
+        (subject, ruling) = disputeManager.computeRuling(_disputeId);
```
sohkai, line 124: Double checking, this will revert if the dispute is not ready yet or if the ruling has already been computed?

packages/evm/contracts/court/AragonCourt.sol
```diff
@@ -86,15 +83,26 @@ contract AragonCourt is Controller, IArbitrator {
...
+    /**
+    * @notice Submit `_evidence` as evidence from `_submitter` for dispute #`_disputeId`
+    * @param _disputeId Id of the dispute in the Court
+    * @param _submitter Address of the account submitting the evidence
+    * @param _evidence Data submitted for the evidence related to the dispute
+    */
+    function submitEvidence(uint256 _disputeId, address _submitter, bytes calldata _evidence) external {
+        IDisputeManager disputeManager = IDisputeManager(_getDisputeManager());
+        (IArbitrable subject ,,,,,) = disputeManager.getDispute(_disputeId);
+        require(subject == IArbitrable(msg.sender), ERROR_SENDER_NOT_DISPUTE_SUBJECT);
```
sohkai, line 102: A bit confused about this; we've moved the `submitEvidence()` interface here, but we would still be expecting users to interact with the arbitrable to submit evidence?

packages/evm/contracts/arbitration/IArbitrable.sol
```diff
@@ -16,37 +12,9 @@ contract IArbitrable is ERC165 {
...
-    /**
-    * @dev Submit evidence for a dispute
-    * @param _disputeId Id of the dispute in the Court
-    * @param _evidence Data submitted for the evidence related to the dispute
-    * @param _finished Whether or not the submitter has finished submitting evidence
-    */
-    function submitEvidence(uint256 _disputeId, bytes calldata _evidence, bool _finished) external;
-
     /**
     * @dev Give a ruling for a certain dispute, the account calling it must have rights to rule on the contract
```
sohkai, line 16: By "rights to rule on the contract", we mean this IArbitrable instance, right?

So we could expect implementations to add their own authentication mechanism to protect this interface?

packages/evm/contracts/test/arbitration/ArbitrableMock.sol
```diff
@@ -5,37 +5,25 @@ import "../../arbitration/IArbitrator.sol";
...
 
-    function rule(uint256 _disputeId, uint256 _ruling) external {
-        emit Ruled(IArbitrator(msg.sender), _disputeId, _ruling);
-    }
-
-    function interfaceID() external pure returns (bytes4) {
-        IArbitrable arbitrable;
-        return arbitrable.submitEvidence.selector ^ arbitrable.rule.selector;
+    function rule(uint256 _disputeId) external {
+        (, uint256 ruling) = arbitrator.rule(_disputeId);
```
sohkai, line 26: If I understand correctly, the motivation behind this is to shift the control flow so a user would initiate rulings from the IArbitrable instance rather than going directly to the IArbitrator (e.g. "pull in the ruling" rather than "push out the ruling")?

And in terms of pending state, if we assume the IArbitrable never calls `arbitrator.rule()`, no other state in the IArbitrator is impacted? It's just forever "yeah, you can call me to fetch your ruling, but I don't care if you do or don't"?

packages/evm/test/disputes/disputes-create.js
```diff
@@ -175,14 +174,15 @@ contract('DisputeManager', () => {
...
-        fakeArbitrable = await FakeArbitrable.new(court.address)
-      })
+        const receipt = await court.createDispute(2, '0xabcd', { from: fakeArbitrable })
 
-      it('reverts', async () => {
-        await assertRevert(fakeArbitrable.createDispute(2, '0x'), ARAGON_COURT_ERRORS.SENDER_NOT_ARBITRABLE)
+        assertAmountOfEvents(receipt, DISPUTE_MANAGER_EVENTS.NEW_DISPUTE, { decodeForAbi: DisputeManager.abi })
+        assertEvent(receipt, DISPUTE_MANAGER_EVENTS.NEW_DISPUTE, { expectedArgs: { disputeId: 0, subject: fakeArbitrable, metadata: '0xabcd' }, decodeForAbi: DisputeManager.abi })
       })
     })
```
sohkai, line 187: Should we add a test here to call the `arbitrator.rule()` function from the EOA account?

packages/evm/test/disputes/disputes-settle-round.js
```diff
@@ -80,13 +80,16 @@ contract('DisputeManager', ([_, drafter, appealMaker, appealTaker, juror500, jur
               assertBn(possibleRulings, 2, 'dispute possible rulings do not match')
               assertBn(finalRuling, expectedFinalRuling, 'dispute final ruling does not match')
 
-              const anotherReceipt = await court.executeRuling(disputeId)
+              const anotherReceipt = await court.rule(disputeId)
```
sohkai, line 83: Related to the comment above, I find this test acts weird. Should you be allowed to call `rule()` for the same disputeId multiple times on the IArbitrator? Are we expecting IArbitrables to implement logic to prevent calling this more than once?

packages/evm/test/disputes/disputes-evidence.js
```diff
@@ -130,14 +129,8 @@ contract('DisputeManager', ([_, juror500, juror1000, juror1500]) => {
     })
 
     context('when the sender is not the arbitrable of the dispute', () => {
-      let fakeArbitrable
-
-      beforeEach('mock non arbitrable', async () => {
-        fakeArbitrable = await FakeArbitrable.new(court.address)
-      })
-
       it('reverts', async () => {
```
sohkai, line 132: Theoretically we should be able to submit evidence with `fakeArbitrable` if we set it as the dispute subject?

sohkai (commented):
As you can tell by the comments, I'm a little bit confused about the motivations behind this and how it impacts our mental model for building IArbitrables.

> Move evidence logging to the Arbitrator

Was this mostly an effort to centralize the logging, so we can pick them all up from a single place? Agree this is much easier.

In terms of thinking about evidence submission, I like this as well as it allows us to place the responsibility of controlling who and when evidence can be submitted wholly onto the IArbitrable. Any interfaces / etc. for the protocol will have to "adapt" to different types of IArbitrables.

> Replace executeRuling from Arbitrator in favor of rule to fetch the ruling for a dispute
> Change rule from Arbitrable so it is not intended to be called from the arbitrator only

Trying to understand the impacts of this a bit more, see comments below. But if I understand correctly, I like it.

> With all these changes we can get rid of ERC165

:heart:

----------

I was also surprised to see not many tests changing—this is because the `ArbitrableMock` still maintains the same interface and most interactions (e.g. evidence submission) are designed to go through it?

packages/evm/contracts/court/AragonCourt.sol
```diff
@@ -86,15 +83,26 @@ contract AragonCourt is Controller, IArbitrator {
     */
     function createDispute(uint256 _possibleRulings, bytes calldata _metadata) external returns (uint256) {
```
sohkai, line 84: ⚠️ Note that at the moment, an arbitrable would not be able to call `rule()` if the DisputeManager module was changed

packages/evm/contracts/arbitration/IArbitrator.sol
facuspagnuolo: I think it's safe to assume it is linked to the dispute

packages/evm/contracts/court/AragonCourt.sol
facuspagnuolo: I will revert if the dispute is not ready to be ruled, but it will simply return the computed ruling if it was already ruled before

packages/evm/contracts/court/AragonCourt.sol
facuspagnuolo: Yes, the main reason is to avoid relying on the Arbitrable to log the Arbitrator correctly. Additionally, we are now being flexible on whatever functionality users would like to provide for submitting evidence on their ends

packages/evm/contracts/court/AragonCourt.sol
facuspagnuolo: Addressed in https://github.com/aragon/protocol/pull/19/commits/8f6c97ac5ee6931a685988520d07d6e393c6c0a3

packages/evm/contracts/test/arbitration/ArbitrableMock.sol
facuspagnuolo: Exactly, this is basically removing the dependency of the Arbitrable in the Arbitrator, now is totally decoupled. The Arbitrable can call `rule` multiple times or not call it at all, the Arbitrator just doesn't care

packages/evm/test/disputes/disputes-create.js
facuspagnuolo: I think we were already testing it [here](https://github.com/aragon/protocol/blob/development/packages/evm/test/disputes/disputes-settle-round.js#L72)

packages/evm/test/disputes/disputes-settle-round.js
facuspagnuolo: > Are we expecting IArbitrables to implement logic to prevent calling this more than once?

No

> Should you be allowed to call rule() for the same disputeId multiple times on the IArbitrator?

Yes, but note that the test is actually ensuring nothing happens (i.e. the event is not re-emitted, and the ruling stays the same)

packages/evm/test/disputes/disputes-evidence.js
facuspagnuolo: Addressed in d042bfc9e25617d667573d7efc824159c25502c9

packages/evm/contracts/arbitration/IArbitrable.sol
facuspagnuolo: > By "rights to rule on the contract", we mean this IArbitrable instance, right?

In the context of the arbitrator yes, but here it doesn't sound very well. Removed in d042bfc9e25617d667573d7efc824159c25502c9

> So we could expect implementations to add their own authentication mechanism to protect this interface?

Yes :)

packages/evm/contracts/arbitration/IArbitrator.sol
facuspagnuolo: Addressed in d042bfc9e25617d667573d7efc824159c25502c9

packages/evm/test/helpers/utils/errors.js
facuspagnuolo: Addressed in d042bfc9e25617d667573d7efc824159c25502c9

packages/evm/test/disputes/disputes-settle-round.js
sohkai: To come back to this, I think we actually want to say "Yes" to this question:

> Are we expecting IArbitrables to implement logic to prevent calling this more than once?

If an IArbitrable is only meant to receive a ruling once (e.g. to unlock some escrowed funds), then it is up to that contract to ensure it does not receive the same ruling twice.

PR #4, opened 2020-10-15
Controller: Allow extending arbitrator interface

facuspagnuolo:
Adding a simple mechanism to allow people registering new functions in the entry point


packages/evm/contracts/test/court/controller/ControlledMock.sol
```diff
@@ -5,10 +5,25 @@ import "../../../court/controller/Controlled.sol";
 
 contract ControlledMock is Controlled {
```
sohkai, line 6: Perhaps the changes here would be worth moving into a different contract, so the other uses of `ControlledMock` don't include this additional logic?

packages/evm/test/court/controller/controller-modules.js
```diff
@@ -616,4 +617,118 @@ contract('Controller', ([_, fundsGovernor, configGovernor, modulesGovernor, some
       })
     })
   })
+
+  describe('customFunctions', () => {
+    let module
+
+    const setCounterSig = sha3('setCounter(uint256)').slice(0, 10)
```
sohkai, line 624: Haha, what do you think about using one of the `web3-eth-abi` encoding methods instead (e.g. [`encodeFunctionSignature()`](https://web3js.readthedocs.io/en/v1.3.0/web3-eth-abi.html#encodefunctionsignature))?

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -45,9 +46,13 @@ contract Controller is IsContract, Modules, CourtClock, CourtConfig {
     // List of all modules registered for the system indexed by address
     mapping (address => Module) internal allModules;
 
+    // List of custom functions indexed by signature
```
sohkai, line 49:
```suggestion
    // List of custom function targets indexed by signature
```

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -129,6 +134,27 @@ contract Controller is IsContract, Modules, CourtClock, CourtConfig {
         _setModulesGovernor(_governors[2]);
     }
 
+    /**
+    * @dev Fallback function allows to forward calls to a specific address in case it was previously registered
+    *      Note the sender will be always the controller in case it is forwarded
+    */
+    function () external payable {
+        address target = customFunctions[msg.sig];
```
sohkai, line 142: Might be worth adding a test for fallback functions; I actually don't know how they would get registered (with `0x00000000`?)

sohkai (commented):
I think I'm missing the context for this particular change, and I'm not entirely sure it should belong with the module governor.

Maybe we can chat about the motivation behind the change and the expected use cases?

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -45,9 +46,13 @@ contract Controller is IsContract, Modules, CourtClock, CourtConfig {
     // List of all modules registered for the system indexed by address
     mapping (address => Module) internal allModules;
 
+    // List of custom functions indexed by signature
+    mapping (bytes4 => address) internal customFunctions;
```
sohkai, line 50: Could we make this public to aid debugging?

packages/evm/test/court/controller/controller-modules.js
```diff
@@ -616,4 +617,118 @@ contract('Controller', ([_, fundsGovernor, configGovernor, modulesGovernor, some
...
+
+        context('when the function was already set', () => {
+          beforeEach('set custom function', async () => {
+            await controller.setCustomFunction(failSig, modulesGovernor, { from })
+            await controller.setCustomFunction(setCounterSig, modulesGovernor, { from })
+            await controller.setCustomFunction(receiveEtherSig, modulesGovernor, { from })
+          })
+
+          itRegistersTheCustomFunction()
+        })
```
sohkai, line 687: Perhaps another test we should add is an attempt at overriding an existing function signature.

packages/evm/contracts/court/controller/Controller.sol
```diff
@@ -234,6 +260,16 @@ contract Controller is IsContract, Modules, CourtClock, CourtConfig {
         _setModulesGovernor(ZERO_ADDRESS);
     }
 
+    /**
+    * @notice Set custom function `_sig` for `_target`
+    * @param _sig Signature of the function to be set
+    * @param _target Address of the target implementation to be registered for the given signature
+    */
+    function setCustomFunction(bytes4 _sig, address _target) external onlyModulesGovernor {
```
sohkai, line 268: Should we add this to the docs, and perhaps explain our intended use case?

packages/evm/test/court/controller/controller-modules.js
```diff
@@ -616,4 +617,118 @@ contract('Controller', ([_, fundsGovernor, configGovernor, modulesGovernor, some
...
+          it('can be called', async () => {
+            const data = module.contract.methods.setCounter(10).encodeABI()
+            await controller.setCustomFunction(setCounterSig, module.address, { from })
+
+            assertBn(await module.counter(), 0, 'counter does not match')
+            await controller.sendTransaction({ data })
+            assertBn(await module.counter(), 10, 'counter does not match')
+          })
+
+          it('handles eth transfers properly', async () => {
```
sohkai, line 656: Could we add a test where we send just ETH or an empty-data transaction? I would suspect `msg.sig` would be `0x00000000`.

facuspagnuolo: Addressed in 3e5d18a6eeb5c9e20f57eacea79ba35151820a98

packages/evm/test/court/controller/controller-modules.js
facuspagnuolo: Addressed in 3e5d18a6eeb5c9e20f57eacea79ba35151820a98

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: Addressed in 3e5d18a6eeb5c9e20f57eacea79ba35151820a98

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: Addressed in 3e5d18a6eeb5c9e20f57eacea79ba35151820a98

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: Addressed in 3e5d18a6eeb5c9e20f57eacea79ba35151820a98

packages/evm/test/court/controller/controller-modules.js
facuspagnuolo: Addressed in 3e5d18a6eeb5c9e20f57eacea79ba35151820a98

packages/evm/contracts/test/court/controller/ControlledMock.sol
facuspagnuolo: Addressed in 3e5d18a6eeb5c9e20f57eacea79ba35151820a98

packages/evm/test/court/controller/controller-modules.js
facuspagnuolo: Addressed in 3e5d18a6eeb5c9e20f57eacea79ba35151820a98

packages/evm/contracts/court/controller/Controller.sol
facuspagnuolo: Addressed in 53fa8d05958a711bd11aee4b3d8de5d7fece9e57

PR #5, opened 2020-10-15
Voting: Implement delegation

facuspagnuolo:
This PR introduces two flavors of delegate voting:
1. Reusable permission persisted
2. One-time signature flow


sohkai:
Replaced by https://github.com/aragon/protocol/pull/14.

PR #6, opened 2020-10-16
Payments: Implement new payments module

facuspagnuolo:


packages/evm/contracts/arbitration/IArbitrator.sol
```diff
@@ -2,6 +2,7 @@ pragma solidity ^0.5.8;
 
 import "./IArbitrable.sol";
 import "../lib/os/ERC20.sol";
+import "../payments/IPaymentsBook.sol";
```
sohkai, line 5: I don't think this is being used in this interface and could be removed

packages/evm/contracts/arbitration/IArbitrator.sol
```diff
@@ -52,11 +53,8 @@ interface IArbitrator {
...
-    * @dev Tell the subscription fees information for a subscriber to be up-to-date
-    * @param _subscriber Address of the account paying the subscription fees for
-    * @return recipient Address where the corresponding subscriptions fees must be transferred to
-    * @return feeToken ERC20 token used for the subscription fees
-    * @return feeAmount Total amount of fees that must be allowed to the recipient
+    * @dev Tell the payments recipient address
+    * @return Address of the payments recipient module
     */
-    function getSubscriptionFees(address _subscriber) external view returns (address recipient, ERC20 feeToken, uint256 feeAmount);
+    function getPaymentsRecipient() external view returns (address);
```
sohkai, line 59: Curious how you'd feel about including variable names in all interface-level functions (don't have to do this now, but in general as a rule).

Sometimes they're obvious, but sometimes the return value can be ambiguous (e.g. `returns (address, address)` or even `createDispute(...) returns (uint256)`).

packages/evm/contracts/court/AragonCourt.sol
```diff
@@ -6,13 +6,12 @@ import "./controller/Controller.sol";
 import "../arbitration/IArbitrator.sol";
 import "../arbitration/IArbitrable.sol";
 import "../disputes/IDisputeManager.sol";
-import "../subscriptions/ISubscriptions.sol";
+import "../payments/IPaymentsBook.sol";
```
sohkai, line 9: Similarly, I don't think this needs to be imported

packages/evm/contracts/court/controller/ControlledRecoverable.sol
```diff
@@ -33,13 +33,22 @@ contract ControlledRecoverable is Controlled {
...
     * @notice Transfer all `_token` tokens to `_to`
-    * @param _token ERC20 token to be recovered
+    * @param _token Address of the token to be recovered
     * @param _to Address of the recipient that will be receive all the funds of the requested token
     */
-    function recoverFunds(ERC20 _token, address _to) external onlyFundsGovernor {
-        uint256 balance = _token.balanceOf(address(this));
-        require(balance > 0, ERROR_INSUFFICIENT_RECOVER_FUNDS);
-        require(_token.safeTransfer(_to, balance), ERROR_RECOVER_TOKEN_FUNDS_FAILED);
+    function recoverFunds(address _token, address payable _to) external payable onlyFundsGovernor {
```
sohkai, line 39: Was this change why the optimization runs were decreased to 100? Seems quite steep just to include an additional condition.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+    // Permyriad of collected fees that will be allocated to the governor of the Court (‱ - 1/10,000)
+    uint16 public governorSharePct;
+
+    // List of periods indexed by ID
+    mapping (uint256 => Period) internal periods;
+
+    event PaymentReceived(uint256 indexed periodId, address indexed payer, address indexed token, uint256 amount, address sender, bytes data);
+    event JurorFeesClaimed(uint256 indexed periodId, address indexed juror, address indexed token, uint256 amount);
+    event GovernorFeesTransferred(uint256 indexed periodId, address indexed token, uint256 amount);
+    event GovernorSharePctChanged(uint16 previousGovernorSharePct, uint16 currentGovernorSharePct);
```
sohkai, line 62: Not a big deal, but I wouldn't mind removing the previous amount here to save gas.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+
+        // Update collected fees for the governor
+        uint256 governorFees = _amount.pct(governorSharePct);
+        period.governorFees[_token] = period.governorFees[_token].add(governorFees);
+
+        // Update collected fees for the jurors
+        uint256 jurorFees = _amount.sub(governorFees);
+        period.jurorFees[_token] = period.jurorFees[_token].add(jurorFees);
+
+        // Deposit tokens from sender to this contract
```
sohkai, line 100: Might be worth leaving a note that `_deposit()` checks `msg.value` on ETH transfers

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+            // solium-disable-next-line security/no-send
+            require(_to.send(_amount), ERROR_ETH_TRANSFER_FAILED);
+        } else {
+            require(ERC20(_token).safeTransfer(_to, _amount), ERROR_TOKEN_TRANSFER_FAILED);
+        }
+    }
+
+    /**
+    * @dev Internal function to make sure that the balance details of a certain period have been computed. This function assumes given ID and
+    *      period correspond to each other.
```
sohkai, line 391: Similar to a note on the voting module, would it also be worth specifying that we assume the given period has already been checked to be in the past?

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+        internal
+        view
+        returns (uint256)
+    {
+        if (_jurorActiveBalance == 0) {
+            return 0;
+        }
+
+        // Note that we already checked the juror active balance is greater than zero.
+        // Then, the total active balance must be greater than zero.
```
sohkai, line 511: We may want to correct this comment; this is more of an assumption that we expect the total active balance _should_ be greater than zero.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+
+    /**
+    * @dev Internal function to transfer tokens
+    * @param _to Recipient of the transfer
+    * @param _token Address of the token to transfer
+    * @param _amount Amount to be transferred
+    */
+    function _transfer(address payable _to, address _token, uint256 _amount) internal {
+        if (_token == address(0)) {
+            // solium-disable-next-line security/no-send
```
sohkai, line 382: We might as well just disable this rule in the config, since it's now an anti-practice to use `.transfer()`.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+    /**
+    * @notice Transfer owed fees to the governor for period #`_periodId`
+    * @param _periodId Identification number of the period being claimed
+    * @param _token Address of the token to be claimed
+    */
+    function transferGovernorFees(uint256 _periodId, address _token) external {
+        require(_periodId <= _getCurrentPeriodId(), ERROR_NON_PAST_PERIOD);
+
+        Period storage period = periods[_periodId];
+        address payable governor = address(uint160(_configGovernor()));
```
sohkai, line 154: Doesn't `_configGovernor()` already return an address?

packages/evm/test/payments/payments-execution.js
```diff
@@ -0,0 +1,144 @@
...
+            it('emits an event', async () => {
+              const currentPeriodId = await paymentsBook.getCurrentPeriodId()
+              const receipt = await paymentsBook.pay(token.address, amount, someone, data, { from })
+
+              assertAmountOfEvents(receipt, PAYMENTS_BOOK_EVENTS.PAYMENT_RECEIVED)
+              assertEvent(receipt, PAYMENTS_BOOK_EVENTS.PAYMENT_RECEIVED, { expectedArgs: { periodId: currentPeriodId, payer: someone, sender: from, token, amount } })
+            })
+          })
+
+          context('when paying with ETH', () => {
```
sohkai, line 90: As noted in the contracts, it may be worth adding a test to revert if you send ETH but don't use the ETH address (along the account already having an approval for specified token so the `transferFrom()` works).

packages/evm/test/payments/payments-initialization.js
```diff
@@ -0,0 +1,123 @@
...
+        })
+      })
+
+      context('when the given governor share is above 100%', () => {
+        const governorSharePct = bn(10001)
+
+        it('reverts', async () => {
+          await assertRevert(PaymentsBook.new(controller.address, PERIOD_DURATION, governorSharePct), PAYMENTS_BOOK_ERRORS.OVERRATED_GOVERNOR_SHARE_PCT)
+        })
+      })
```
sohkai, line 63: As noted below, it may be worth adding a test that fails with too large a period duration

packages/evm/test/payments/payments-distribution.js
```diff
@@ -0,0 +1,437 @@
...
+
+            it('ignores duplicated governor requests', async () => {
+              const previousBalance = await token.balanceOf(governor)
+
+              await paymentsBook.transferGovernorFees(periodId, token.address)
+              await paymentsBook.transferGovernorFees(periodId, token.address)
+
+              const currentBalance = await token.balanceOf(governor)
+              assertBn(currentBalance, previousBalance.add(expectedGovernorTokenFees), 'governor token balance does not match')
+            })
```
sohkai, line 217: Might be worth double checking that the second call doesn't emit any events

packages/evm/test/payments/payments-distribution.js
```diff
@@ -0,0 +1,437 @@
...
+    })
+
+    context('when there were no payments', () => {
+      context('when requesting a past period', () => {
+        const periodId = 0
+
+        it('ignores the request', async () => {
+          const previousBalance = await token.balanceOf(paymentsBook.address)
+
+          await paymentsBook.claimJurorFees(periodId, token.address, { from: jurorPeriod0Term1 })
```
sohkai, line 413: Similarly, may be worth double checking no events are emitted.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+    /**
+    * @dev Internal function to transfer governor fees for a certain period
+    * @param _period Period being claimed
+    * @param _periodId Identification number of the period being claimed
+    * @param _token Address of the token to be claimed
+    */
+    function _transferGovernorFees(Period storage _period, uint256 _periodId, address payable _governor, address _token) internal {
+        uint256 amount = _getGovernorFees(_period, _token);
+        if (amount > 0) {
+            _period.governorFees[_token] = 0;
```
sohkai, line 354: While this works fabulously, a qualm I have about this is that it overwrites history for the governor's share in a way that does not happen for the guardian's share.

I'm wondering if we could repurpuse the `claimedGuardians` mapping such that it's also used for the governor?

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+    * @dev Internal function to pull tokens into this contract
+    * @param _from Owner of the deposited funds
+    * @param _token Address of the token to deposit
+    * @param _amount Amount to be deposited
+    */
+    function _deposit(address _from, address _token, uint256 _amount) internal {
+        if (_token == address(0)) {
+            require(msg.value == _amount, ERROR_ETH_DEPOSIT_MISMATCH);
+        } else {
+            require(ERC20(_token).safeTransferFrom(_from, address(this), _amount), ERROR_TOKEN_DEPOSIT_FAILED);
```
sohkai, line 370: ⚠️ we should revert here if `msg.value > 0` (or alternatively, switch the condition so we use `msg.value` as the first condition and revert if `_token != address(0)`  inside the ETH branch).

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    /**
+    * @dev Internal function to get the Court term in which a certain period starts
+    * @param _periodId Identification number of the period querying the start term of
+    * @return Court term where the given period starts
+    */
+    function _getPeriodStartTermId(uint256 _periodId) internal view returns (uint64) {
+        // Periods are measured in Court terms. Since Court terms are represented in uint64, we are safe to use uint64 for period ids too.
+        // We are using SafeMath here because if any user calls `getPeriodBalanceDetails` for a huge period ID,
```
sohkai, line 472: I don't think this comment is still valid; `getPeriodBalanceDetails()` doesn't call `_getPeriodStartTermId()`.

However, while it's not that big of a problem, you can technically cause this to always fail if you set a large enough `periodDuration` on initialization (e.g. UINT64_MAX). One way around this is to double check a multiplication against the value, e.g. 1000k, works when setting the period duration.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    /**
+    * @dev Check if a given juror has already claimed the owed fees for a certain period
+    * @param _periodId Identification number of the period being queried
+    * @param _juror Address of the juror being queried
+    * @param _token Address of the token to be queried
+    * @return True if the juror has already claimed the corresponding token fees
+    */
+    function hasJurorClaimed(uint256 _periodId, address _juror, address _token) external view returns (bool) {
```
sohkai, line 282: Curious if what you think about flipping this around to `canGuardianClaim()`?

Then we could also check if the specified guardian has any tokens to claim before answering yes/no, which could make it easier to others to know if they should call the claim.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+    * @dev Internal function to claim juror fees for a certain period
+    * @param _period Period being claimed
+    * @param _periodId Identification number of the period claiming fees for
+    * @param _juror Address of the juror claiming the fees
+    * @param _token Address of the token being claimed
+    * @param _amount Amount of tokens to be transferred to the juror
+    */
+    function _claimJurorFees(Period storage _period, uint256 _periodId, address payable _juror, address _token, uint256 _amount) internal {
+        if (_amount > 0) {
+            _period.claimedJurorFees[_juror][_token] = true;
```
sohkai, line 339: I'm not against this, as it saves gas, but we might consider moving this out of the `if` (or flip the `hasJurorClaimed()` as suggested above).

Otherwise, if you have nothing to claim, you could keep calling (or getting fooled into calling) functions relying on this check haha.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+        _claimJurorFees(period, _periodId, msg.sender, _token, amount);
+    }
+
+    /**
+    * @notice Claim juror fees for period #`_periodId` owed to `msg.sender`
+    * @dev It will ignore tokens that were already claimed without reverting
+    * @param _periodId Identification number of the period which fees are claimed for
+    * @param _tokens List of token addresses to be claimed
+    */
+    function claimManyJurorFees(uint256 _periodId, address[] calldata _tokens) external {
```
sohkai, line 128: I'm wondering if:

1. `claimGuardianShare()` and `transferGovernorShare()` are both redundant, since there are array versions available (and the gas overhead is probably only in the hundreds of gas, if even that)
2. Add the ability to claim by signature. This would let us build additional meta-delegation wrappers as necessary to bundle the txs for multiple periods.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    /**
+    * @dev Tell the fees corresponding to a juror for a certain period
+    * @param _periodId Identification number of the period being queried
+    * @param _juror Address of the juror querying the owed fees of
+    * @param _tokens List of token addresses to be queried
+    * @return List of token amounts corresponding to the juror
+    */
+    function getManyJurorFees(uint256 _periodId, address _juror, address[] calldata _tokens) external view returns (uint256[] memory amounts) {
```
sohkai, line 261: I'm not against these "many" variants, but similar to before, I'm curious how much we want them.

On one side, it's a slight UX gain / gas gain if we anticipate another on-chain contract to call the version with just a single token. On the other hand, we don't necessarily need the "many" version for off-chain interactions, as one could do a multicall (or just multiple `eth_call`s).

At the moment, similar to before, I'm leaning towards simplifying to only keeping the "many" version. Are there any potential problems with subgraph logic if we do this, or is it already just based on the `PaymentReceived` events?

sohkai (commented):
I've left a couple of small suggestions or quick questions at the beginning, and moved most of the code or logic-related comments to the end.

The biggest ones I have in my mind:

- Double checking that when `msg.value > 0`, the `token` lines up
- Design question on the "many" variants
- Whether we should add the ability to claim by signature

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: I agree, addressed in 51b7ed3c9324e62bd2b9bc9b85047db0c062ac7b

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Yes but it is not payable, am I missing sth?

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Addressed in b0ee119be0ba5b590fc137bec28df4dfc05bf4b6

packages/evm/test/payments/payments-execution.js
facuspagnuolo: Addressed in b0ee119be0ba5b590fc137bec28df4dfc05bf4b6

packages/evm/contracts/arbitration/IArbitrator.sol
facuspagnuolo: Addressed in 023212528dd6ef0b83a7411882ebc56b7f310c00

packages/evm/contracts/court/AragonCourt.sol
facuspagnuolo: Addressed in 023212528dd6ef0b83a7411882ebc56b7f310c00

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: I still want to go over all the emitted logs and fix all the unnecessary logged info :)

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Addressed in ee7d2b6144d821812d9ef6ac2f7451288238a17a

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Addressed in ee7d2b6144d821812d9ef6ac2f7451288238a17a

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Addressed in ee7d2b6144d821812d9ef6ac2f7451288238a17a

packages/evm/contracts/arbitration/IArbitrator.sol
facuspagnuolo: Sounds interesting, we can try to cover that entirely once we finish implementing all these feature-related changes

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: I actually did both, moved this out of the if and change `hasGuardianClaimed` for `canGuardianClaim`. Addressed in 55bcbf4056e85ce958a64136d44ad7ea0038c304

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Addressed in 55bcbf4056e85ce958a64136d44ad7ea0038c304

packages/evm/test/payments/payments-distribution.js
facuspagnuolo: Addressed in b3b4648025045022dd2b426518dc3c6f756bc195

packages/evm/test/payments/payments-distribution.js
facuspagnuolo: Addressed in b3b4648025045022dd2b426518dc3c6f756bc195

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: I'll update the comment but I don't it's extremely necessary to have a check for the `periodDuration`, whatever number you use as a benchmark for the multiplication will sound like you cannot have more than that number of periods, which at the end will always depend on the term duration and period duration you're using

packages/evm/test/payments/payments-initialization.js
facuspagnuolo: Answered below

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Addressed in b91fd72bb983dfc0a5a9286a861da01b5cfd6a47

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Addressed in b91fd72bb983dfc0a5a9286a861da01b5cfd6a47

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Addressed in 41d17890c03d2627e2baf44fd7c23843e9d10787

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: Addressed in 51b7ed3c9324e62bd2b9bc9b85047db0c062ac7b and e6fe18fc83ecf84deecb75da5e5f13ab3b76b411 respectively

packages/evm/contracts/arbitration/IArbitrator.sol
sohkai: Let's move this to an issue: https://github.com/aragon/protocol/issues/35 :)

packages/evm/contracts/payments/PaymentsBook.sol
sohkai: Nevermind me, just realizing now that in 0.5 we have to cast back to an `uint160` first!

packages/evm/contracts/payments/PaymentsBook.sol
sohkai: Right, that makes sense. You'll never be able to claim shares, but that's because the first period hasn't ended (and won't ever end).

PR #9, opened 2020-10-20
Deployments: Create deployment script

facuspagnuolo:


packages/deployment/scripts/1-deploy-protocol/input/protocol.js
```diff
@@ -0,0 +1,13 @@
+const localhost = require('./protocol.localhost')
```
sohkai, line 1: To not be confused with the protocol configs per environment, should we rename this file as `index.js`?

In the future, if we have multiple input configs per environment, we could then move the exports here to be mappings to individual configs, e.g.

```
const ropsten = {
  protocol: require('./protocol.ropsten')
}
```

packages/deployment/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "deploy-faucet": "buidler deploy-faucet"
+  },
+  "dependencies": {
+    "@aragon/protocol-evm": "^1.0.0",
+    "@aragon/contract-helpers-test": "^0.1.0",
+    "@aragonone/erc20-faucet": "https://github.com/aragonone/erc20-faucet",
+    "@nomiclabs/buidler": "^1.4.3",
+    "@nomiclabs/buidler-truffle5": "^1.3.4",
+    "@nomiclabs/buidler-web3": "^1.3.4",
+    "@truffle/contract": "4.0.32",
```
sohkai, line 21: Haha this is interesting, are we not able to use `artifacts.require()` directly?

packages/deployment/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "author": "Aragon Association <legal@aragon.org>",
+  "license": "GPL-3.0",
+  "scripts": {
+    "help": "buidler help",
+    "node": "buidler node",
+    "console": "buidler console",
+    "deploy-protocol": "buidler deploy-protocol",
+    "deploy-faucet": "buidler deploy-faucet"
+  },
+  "dependencies": {
```
sohkai, line 14: Nitpicking, but theoretically we could mark this package as private and move all of these dependencies as dev dependencies?

packages/deployment/README.md
```diff
@@ -0,0 +1,38 @@
+# Aragon Protocol deployment scripts
+
+This package includes the deployment scripts used for Aragon Protocol 
+
+## Installation
+
+Clone this monorepo and install dependencies:
+
+```sh
+git clone git@github.com:aragon/protocol.git
```
sohkai, line 10: Might be worth adding a `cd` in case it's not obvious :).

packages/deployment/scripts/1-deploy-protocol/output/protocol.rinkeby.json
```diff
@@ -0,0 +1,26 @@
+{
```
sohkai, line 1: We don't have to do this now (let's leave it as an issue for when we do mainnet), but what do you think about re-deploying the "official" Rinkeby and Ropsten instances such that all the token and module addresses are the same?

packages/deployment/scripts/2-deploy-faucet/input/faucet.js
```diff
@@ -0,0 +1,66 @@
+const { bigExp } = require('@aragon/contract-helpers-test')
```
sohkai, line 1: Similar to the comment above on the protocol, do you think this fits better as an `index.js`?

packages/deployment/src/shared/Artifacts.js
```diff
@@ -0,0 +1,39 @@
+const fs = require('fs')
```
sohkai, line 1: Just wondering, can't we use the bre to get the truffle artifact directly? Or is this not possible because we're in a different project namespace and we're not requiring buidler to compile anything?

packages/deployment/src/deployers/ERC20Deployer.js
```diff
@@ -0,0 +1,40 @@
+const BaseDeployer = require('./BaseDeployer')
+const logger = require('../helpers/logger')('ERC20Deployer')
+
+module.exports = class extends BaseDeployer {
+  constructor(config, environment, output) {
+    super(environment, output)
+    this.config = config
+  }
+
+  async call() {
```
sohkai, line 10: Maybe this is better named as `deploy()` 😄

packages/deployment/src/deployers/FaucetDeployer.js
```diff
@@ -0,0 +1,68 @@
...
+    this.config = config
+  }
+
+  async call() {
+    const address = this.previousDeploy.address
+    const ERC20Faucet = await this.environment.getArtifact('ERC20Faucet', '@aragonone/erc20-faucet')
+
+    return address
+        ? this._loadFaucet(ERC20Faucet, address)
+        : this._deployFaucet(ERC20Faucet)
```
sohkai, line 18: Tiny nitpick on spacing

packages/deployment/src/deployers/FaucetDeployer.js
```diff
@@ -0,0 +1,68 @@
...
+    logger.warn(`Using previous deployed faucet instance at ${address}`)
+    return ERC20Faucet.at(address)
+  }
+
+  async _deployFaucet(ERC20Faucet) {
+    const sender = await this.environment.getSender()
+    this._printFaucetDeploy(sender)
+    const { owner, tokens } = this.config
+
+    const tokensAddresses = tokens.map(token => token.address)
```
sohkai, line 31: Wondering if we should use the `ERC20Deployer` here as well? I don't really foresee a case where you'd want to deploy the faucet first, but it might be useful to either:

- Document that `deploy-faucet` is dependent on a previous protocol deploy, or
- Allow this to happen on its own

packages/deployment/src/deployers/ProtocolDeployer.js
```diff
@@ -0,0 +1,298 @@
...
+    const sender = await this.environment.getSender()
+    const modulesGovernor = await this.protocol.getModulesGovernor()
+
+    if (modulesGovernor === sender) {
+      logger.info('Setting modules...')
+      const ids = [MODULES.DISPUTE_MANAGER, MODULES.TREASURY, MODULES.VOTING, MODULES.GUARDIANS_REGISTRY, MODULES.PAYMENTS_BOOK]
+      const implementations = [this.disputes, this.treasury, this.voting, this.registry, this.paymentsBook].map(i => i.address)
+      await this.protocol.setModules(ids, implementations)
+      logger.info('Caching modules...')
+      await this.protocol.cacheModules(implementations, ids)
```
sohkai, line 103: We've made a few changes to the module system, so perhaps this can be done in the same call now :)

packages/deployment/src/deployers/ERC20Deployer.js
```diff
@@ -0,0 +1,40 @@
...
+    const token = this.previousDeploy[this.config.symbol]
+    const ERC20 = await this.environment.getArtifact('ERC20Mock', '@aragon/protocol-evm')
```
sohkai, line 12: We can leave another issue for this, but it might be worth adding a mock of ANTv2 so that the test network's tokens have the same `permit()` and `transferWithAuth()` functionalities.

sohkai (commented):
I didn't get a chance to finish this review before, but let me know what you think and let's move anything highlighted to issues that we can solve before deployment :)

PR #12, opened 2020-10-22
Subgraph: Update protocol subgraph

facuspagnuolo:


packages/deployment/package.json
```diff
@@ -23,6 +22,6 @@
     "web3": "^1.2.11"
   },
   "devDependencies": {
-    "@nomiclabs/buidler-ganache": "^1.3.3"
+    "ganache-cli": "^6.11.0"
```
sohkai, line 25: Curious behind the switch, was it to get to a newer ganache-core version or was the plugin just not starting ganache correctly?

scripts/start.sh
```diff
@@ -0,0 +1,39 @@
...
+  echo "Deploying Aragon Protocol to local node..."
+  yarn workspace @aragon/protocol-deployment deploy-protocol --network ganache
+}
+
+start_local_node
+deploy_protocol
+
+while true; do
+  read
+done
```
sohkai, line 39: Is this just to keep the ganache process running?

packages/subgraph/subgraph.template.yaml
```diff
@@ -1,36 +1,35 @@
...
+        - Evidence
       abis:
-        - name: AragonCourt
-          file: ./node_modules/@aragon/court/abi/AragonCourt.json
-        - name: JurorsRegistry
-          file: ./node_modules/@aragon/court/abi/JurorsRegistry.json
-        - name: Subscriptions
-          file: ./abi/CourtSubscriptions.json
+        - name: AragonProtocol
+          file: ./node_modules/@aragon/protocol-evm/abi/AragonProtocol.json
```
sohkai, line 26: Now that this is a shared workspace, I just wanted to double check:

1. Are we expecting the subgraph package to have the `protocol-evm` package linked?
2. If yes to the above (I'd agree with "yes"), we should make sure to compile the `protocol-evm` package. At the moment, I believe this isn't done automatically and so these links to the compiled ABI will fail.

packages/subgraph/src/Voting.ts
```diff
@@ -1,25 +1,27 @@
...
+
+import { Controller } from '../types/templates/GuardiansRegistry/Controller'
+import { GuardianDraft, Vote } from '../types/schema'
+import { Voting, VoteCommitted, VoteLeaked, VoteRevealed } from '../types/templates/Voting/Voting'
 
 export function handleVoteCommitted(event: VoteCommitted): void {
-  let disputeRoundId = event.params.voteId
-  let draftId = buildDraftId(disputeRoundId, event.params.voter)
-  let draft = loadOrCreateJurorDraft(draftId, disputeRoundId, event.params.voter, event)
+  const disputeRoundId = event.params.voteId
```
sohkai, line 10: TBH I thought `const` wasn't available in TheGraph's version of AssemblyScript. Is this new or did we just not know about it before?

packages/subgraph/schema.graphql
```diff
@@ -24,47 +24,40 @@ type CourtConfig @entity {
...
+  randomnessBN: BigInt!
+  randomness: Bytes
+  protocol: Protocol!
+  createdAt: BigInt!
 }
 
 type Dispute @entity {
   id: ID!
-  subject: Arbitrable!
+  subject: Bytes!
```
sohkai, line 60: Maybe this is something for later, but wouldn't we want to keep the link to an Arbitrable schema? Or is this considered unnecessary because of the new evidence submission process?

packages/subgraph/schema.graphql
```diff
@@ -24,47 +24,40 @@ type CourtConfig @entity {
...
+type ProtocolModule @entity {
   id: ID!
-  startTime: BigInt!
-  randomnessBN: BigInt!
-  randomness: Bytes
-  court: CourtConfig!
-  createdAt: BigInt!
+  protocol: Protocol!
+  moduleId: String!
+  type: ProtocolModuleType!
```
sohkai, line 37: We can do this later, but perhaps it's worth also noting down the cached modules in each module?

packages/subgraph/schema.graphql
```diff
@@ -24,47 +24,40 @@ type CourtConfig @entity {
...
   configGovernor: Bytes
   modulesGovernor: Bytes
-  jurorsRegistry: JurorsRegistryModule @derivedFrom(field: "court")
-  subscriptions: SubscriptionModule @derivedFrom(field: "court")
-  modules: [CourtModule!] @derivedFrom(field: "court")
-  terms: [CourtTerm!] @derivedFrom(field: "court")
-  moduleAddresses: [String!]!
+  guardiansRegistry: GuardiansRegistryModule @derivedFrom(field: "protocol")
+  paymentsBook: PaymentsBookModule @derivedFrom(field: "protocol")
+  modules: [ProtocolModule!] @derivedFrom(field: "protocol")
```
sohkai, line 29: Does this field contain just the currently active modules or all historical modules (don't think this is that important at the moment)?

packages/subgraph/src/PaymentsBook.ts
```diff
@@ -0,0 +1,105 @@
...
+  shareClaim.save()
+}
+
+export function handleGovernorSharePctChanged(event: GovernorSharePctChanged): void {
+  const module = loadOrCreatePaymentsBookModule(event.address)
+  module.governorSharePct = BigInt.fromI32(event.params.currentGovernorSharePct)
+  module.save()
+}
+
+export function updateCurrentPaymentPeriod(address: Address, timestamp: BigInt, token: Address | null = null): void {
```
sohkai, line 46: Could the `token` be null here?

packages/subgraph/src/Voting.ts
```diff
@@ -39,37 +41,37 @@ export function handleVoteRevealed(event: VoteRevealed): void {
...
+    const voting = Voting.bind(event.address)
+    const controllerAddress = voting.getController()
+    const controller = Controller.bind(controllerAddress)
+    const disputeManagerAddress = controller.getDisputeManager().value0
+    const disputeRoundIdArray = decodeDisputeRoundId(disputeRoundId)
+    draft = createGuardianDraft(disputeManagerAddress, disputeRoundIdArray[0], disputeRoundIdArray[1], guardianAddress, event.block.timestamp)
   }
 
-  return draft
+  return draft!
```
sohkai, line 74: Double checking, in all these cases where we're removing `null` as a potential return type, it's because we've already ensured that the entity was "load or created", and are just cleaning up the types now?

sohkai (commented):
A couple of things I noticed; I'd only consider the one about automatically compiling the `protocol-evm` package to be important so it's easier to set up the subgraph environment.

Otherwise, I think a bunch of other things could be lifted into issues for later if you'd prefer!

scripts/start.sh
facuspagnuolo: Yup

packages/deployment/package.json
facuspagnuolo: Nope. We weren't using the ganache plugin for buidler actually, I just needed to start a ganache instance. Is there a way to do that from the plugin? I thought it was simply wrapping some executions in a ganache instance but I don't think it exposes an interface to control the ganache instance yourself

packages/subgraph/subgraph.template.yaml
facuspagnuolo: 1. Yes
2. I tried using the "prepare" script or sth similar in `protocol-evm` but it started making GH actions fail cause it couldn't download the compiler version properly :/

I'll try adding sth in the subgraph installation step instead

packages/subgraph/schema.graphql
facuspagnuolo: Exactly, we were using it for the evidence which is now registered by the arbitrator and linked in the dispute directly

packages/subgraph/schema.graphql
facuspagnuolo: Yes definitely saving this for later

packages/subgraph/src/Voting.ts
facuspagnuolo: It wasn't but it is just being warned now, I found it more appropriate tbh

packages/subgraph/schema.graphql
facuspagnuolo: Currently all the historical modules (indexed by address), we will need to implement some of the new events to actually have this listed by active/inactive and their custom functions

packages/subgraph/src/PaymentsBook.ts
facuspagnuolo: Yes, based on whether there is a token we are updating the payments corresponding to that token or not. There are some events like a heartbeat that doesn't require updating payments related info

packages/subgraph/src/Voting.ts
facuspagnuolo: Yes, not sure if I missed the `!` functionality before of it's new tbh, but I found it useful :)

packages/subgraph/subgraph.template.yaml
facuspagnuolo: Addressed in 66107356030fa181d64da310d82d17d928e13566

packages/deployment/package.json
sohkai: Not sure if this addresses it, but you can configure the ganache-core instance from the buidler plugin, [see docs](https://hardhat.org/plugins/nomiclabs-hardhat-ganache.html#configuration).

I think it will keep it open as well if you use `npx buidler node --network ganache`, so we don't have to do the `while true;` below.

packages/subgraph/schema.graphql
sohkai: Made an issue: https://github.com/aragon/protocol/issues/23!

packages/subgraph/schema.graphql
sohkai: In that case, maybe we should name it `historicalModules` or `allModules` instead. But I don't think we need to rely on this at the moment and let's pull it into https://github.com/aragon/protocol/issues/23.

PR #14, opened 2020-10-22
Voting: Implement delegation

facuspagnuolo:
*Restoring wrongly merged #5*

This PR introduces two flavors of delegate voting:
1. Reusable permission persisted
2. One-time signature flow


packages/evm/test/helpers/utils/crvoting.js
```diff
@@ -1,5 +1,6 @@
+const abi = require('web3-eth-abi')
 const { bn } = require('@aragon/contract-helpers-test')
-const { soliditySha3 } = require('web3-utils')
+const { keccak256, soliditySha3 } = require('web3-utils')
```
sohkai, line 3: Noticed we use `sha3` and `keccak256` interchangeably a few times in different files. Should we align on `keccak256`?

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -44,11 +56,14 @@ contract CRVoting is Controlled, ICRVoting {
 
     // Vote records indexed by their ID
     mapping (uint256 => Vote) internal voteRecords;
+    // Representatives indexed by principals
+    mapping (address => mapping (address => bool)) representatives;
```
sohkai, line 60:
```
    mapping (address => mapping (address => bool)) internal representatives;
```

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -91,12 +120,80 @@ contract CRVoting is Controlled, ICRVoting {
...
+    * @notice Commit multiple votes
+    * @param _voteIds List of IDs of the vote instances to commit a vote to
+    * @param _voters List of addresses of the voters to commit a vote for
+    * @param _commitments List of hashed outcomes to be stored for future reveals
+    */
+    function commitForMany(uint256[] calldata _voteIds, address[] calldata _voters, bytes32[] calldata _commitments)
+        external
+    {
+        require(_voteIds.length == _voters.length, ERROR_INVALID_INPUTS_LENGTH);
+        require(_voteIds.length == _commitments.length, ERROR_INVALID_INPUTS_LENGTH);
```
sohkai, line 161: Maybe it's a little bit more efficient to do an `&&` rather than two requires

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -91,12 +120,80 @@ contract CRVoting is Controlled, ICRVoting {
...
+    function commitFor(uint256 _voteId, address _voter, bytes32 _commitment) external voteExists(_voteId) {
+        require(_isRepresentativeOf(_voter, msg.sender), ERROR_SENDER_NOT_REPRESENTATIVE);
+        _commitFor(_voteId, _voter, _commitment);
+    }
+
+    /**
+    * @notice Commit a vote on behalf of `_voter` for vote #`_voteId`
+    * @param _voteId ID of the vote instance to commit a vote to
+    * @param _voter Address of the voter to commit a vote for
+    * @param _commitment Hashed outcome to be stored for future reveal
```
sohkai, line 141: We should add the v, r, and s params to natspec

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -91,12 +120,80 @@ contract CRVoting is Controlled, ICRVoting {
...
+    * @param _commitments List of hashed outcomes to be stored for future reveals
+    */
+    function commitForMany(uint256[] calldata _voteIds, address[] calldata _voters, bytes32[] calldata _commitments)
+        external
+    {
+        require(_voteIds.length == _voters.length, ERROR_INVALID_INPUTS_LENGTH);
+        require(_voteIds.length == _commitments.length, ERROR_INVALID_INPUTS_LENGTH);
+
+        for (uint256 i = 0; i < _voteIds.length; i++) {
+            require(_existsVote(voteRecords[_voteIds[i]]), ERROR_VOTE_DOES_NOT_EXIST);
```
sohkai, line 164: This is not particularly related to this PR, but what do you think about having `_existsVote()` take in a vote id rather than vote storage instance? As far as I could tell, all usages were first dereferencing the vote id anyway.

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -91,12 +120,80 @@ contract CRVoting is Controlled, ICRVoting {
...
+            require(_isRepresentativeOf(_voters[i], msg.sender), ERROR_SENDER_NOT_REPRESENTATIVE);
+            _commitFor(_voteIds[i], _voters[i], _commitments[i]);
+        }
+    }
+
+    /**
+    * @notice Commit multiple votes
+    * @param _voteIds List of IDs of the vote instances to commit a vote to
+    * @param _voters List of addresses of the voters to commit a vote for
+    * @param _commitments List of hashed outcomes to be stored for future reveals
```
sohkai, line 174: Reminder on documenting the v, r, and s :)

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -137,6 +234,38 @@ contract CRVoting is Controlled, ICRVoting {
...
+    */
+    function isRepresentativeOf(address _voter, address _representative) external view returns (bool) {
+        return _isRepresentativeOf(_voter, _representative);
+    }
+
+    /**
+    * @dev Verify if a representative address has been authorized by a voter for a specific vote
+    * @param _voteId Identification number of the vote to be checked
+    * @param _voter Address of the voter to be checked
+    * @param _representative Address of the representative to be checked
```
sohkai, line 251: Reminder on documenting the v, r, and s :)

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -137,6 +234,38 @@ contract CRVoting is Controlled, ICRVoting {
...
+    * @param _voter Address of the principal being queried
+    * @param _representative Address of the representative being queried
+    * @return True if the representative currently represents the voter
+    */
+    function isRepresentativeOf(address _voter, address _representative) external view returns (bool) {
+        return _isRepresentativeOf(_voter, _representative);
+    }
+
+    /**
+    * @dev Verify if a representative address has been authorized by a voter for a specific vote
```
sohkai, line 248:
```suggestion
    * @dev Verify if a representative address has been authorized by a voter for a specific vote by signature
```

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -237,6 +373,21 @@ contract CRVoting is Controlled, ICRVoting {
         return _hashVote(_outcome, _salt);
     }
 
+    /**
+    * @dev Commit a vote on behalf of `_voter` for vote #`_voteId`
```
sohkai, line 377: Not sure if we do this in other places in these contracts, but would it be worth mentioning here that we assume the caller has authenticated the args already?

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -301,6 +452,49 @@ contract CRVoting is Controlled, ICRVoting {
...
+    */
+    function _isRepresentativeOf(address _voter, address _representative) internal view returns (bool) {
+        return representatives[_voter][_representative] || _voter == _representative;
+    }
+
+    /**
+    * @dev Verify if a representative address has been authorized by a voter for a specific vote
+    * @param _voteId Identification number of the vote to be checked
+    * @param _voter Address of the voter to be checked
+    * @param _representative Address of the representative to be checked
```
sohkai, line 469: Reminder on documenting the v, r, and s :)

packages/evm/contracts/voting/CRVoting.sol
```diff
@@ -301,6 +452,49 @@ contract CRVoting is Controlled, ICRVoting {
...
+    * @param _representative Address of the representative being queried
+    * @return True if the representative currently represents the voter
+    */
+    function _isRepresentativeOf(address _voter, address _representative) internal view returns (bool) {
+        return representatives[_voter][_representative] || _voter == _representative;
+    }
+
+    /**
+    * @dev Verify if a representative address has been authorized by a voter for a specific vote
+    * @param _voteId Identification number of the vote to be checked
```
sohkai, line 467:
```suggestion
    * @param _voteId ID of the vote instance to be checked
```

packages/evm/test/voting/crvoting-commit.js
```diff
@@ -1,15 +1,16 @@
+const { ecsign } = require('ethereumjs-util')
 const { bn } = require('@aragon/contract-helpers-test')
 const { assertRevert, assertBn, assertAmountOfEvents, assertEvent } = require('@aragon/contract-helpers-test/src/asserts')
 
 const { buildHelper } = require('../helpers/wrappers/protocol')
 const { VOTING_EVENTS } = require('../helpers/utils/events')
-const { OUTCOMES, hashVote } = require('../helpers/utils/crvoting')
+const { OUTCOMES, hashVote, createRepresetativeAuthorization } = require('../helpers/utils/crvoting')
```
sohkai, line 7:
```suggestion
const { OUTCOMES, hashVote, createRepresentativeAuthorization } = require('../helpers/utils/crvoting')
```

packages/evm/test/voting/crvoting-commit.js
```diff
@@ -163,4 +171,439 @@ contract('CRVoting', ([_, voter]) => {
...
+      })
+    })
+  })
+
+  describe('commitForWithSig', () => {
+    let voter, voterPK
+    const from = representative
+
+    before('create wallet', async () => {
+      const wallet = web3.eth.accounts.create('erc3009')
```
sohkai, line 377: Haha, maybe we can update this to `create('commitWithSig')` or something :)

sohkai (commented):
I've left a couple of lesser notes below, but I wanted to bring up our offline discussion about the auth delegation through signatures.

We were discussing `commitForWithSig()` / `commitForManyWithSig()`:

- I think this is useful; this allows us to turn on a service that subsidizes gas fees in the future without having users trust us "forever"
- I am not sure about encoding the `representative` directly in the authorization; it seems unnecessary because as a user, you may either not know who that account is (and as a service provider, we may not want to bind ourselves to one particular account either!), and you don't care who does the commitment for you, as long as it's committed
- Going off of the above, if we let anyone submit commit authorizations, `commitForManyWithSig()` becomes grief-able by bad actors frontrunning a single signature to fail the larger transaction. What do you think about letting `commitForManyWithSig()` handle failures from an authorization being already committed?

We were also discussing whether the "forever" delegation should be allowed via signature:

- I still think this would be quite useful, especially in the context of a cold/hot wallet combination
- We could use the now-common "delegateBySig" functionality / typehash for this

packages/evm/test/helpers/utils/crvoting.js
facuspagnuolo: Addressed in b22857c8c1a036f7be192235139dbf687c072a79

packages/evm/contracts/voting/CRVoting.sol
facuspagnuolo: Addressed in 670f508204bc2cfbfac38ff5327091972e42253b

packages/evm/contracts/voting/CRVoting.sol
facuspagnuolo: Addressed in 6bba69e8432348d34597c4936955e2301bfd9e73

packages/evm/contracts/voting/CRVoting.sol
facuspagnuolo: Fixed in ff4b971e5307ec492cf063eeaa94c4e32ae9ad4f

packages/evm/test/voting/crvoting-commit.js
facuspagnuolo: Fixed in d7e01f6be6c9eaeab8b117e4bff5e64db07a46ce

facuspagnuolo:
@sohkai thx for the review, as we synced offline, I marked as resolved many a few comments that don't apply anymore due to the changes we did in #27, the rest of them were addressed in that PR.

PR #15, opened 2020-10-22
Registry: Implement activation locks

facuspagnuolo:


packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -110,6 +123,8 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
     event GuardianDeactivationRequested(address indexed guardian, uint64 availableTermId, uint256 amount);
     event GuardianDeactivationProcessed(address indexed guardian, uint64 availableTermId, uint256 amount, uint64 processedTermId);
     event GuardianDeactivationUpdated(address indexed guardian, uint64 availableTermId, uint256 amount, uint64 updateTermId);
+    event GuardianDeactivationLockIncreased(address indexed guardian, address indexed lockManager, uint256 amount, uint256 total);
+    event GuardianDeactivationLockDecreased(address indexed guardian, address indexed lockManager, uint256 amount, uint256 total);
```
sohkai: Wonder if we can just use a single Changed event?

packages/evm/contracts/registry/ILockManager.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.5.17;
+
+import "../lib/os/ERC20.sol";
```
sohkai: AFAIK we don't need to include this import here?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -63,9 +67,18 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
...
         uint256 availableBalance;                   // Available tokens that can be withdrawn at any time
         uint64 withdrawalsLockTermId;               // Term ID until which the guardian's withdrawals will be locked
+        DeactivationLocks deactivationLocks;        // Guardian's deactivation locks
         DeactivationRequest deactivationRequest;    // Guardian's pending deactivation request
     }
 
+    /**
+    * @dev Guardians can define lock managers to prevent them from deactivating tokens from the registry
+    */
+    struct DeactivationLocks {
```
sohkai: I wonder if we should re-orient this concept as an "activation lock" instead? Sounds a bit weird to base on it on deactivation since you're committing to keep an amount locked via the manager.

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -164,6 +179,10 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
         uint256 minActiveBalance = _getMinActiveBalance(termId);
         require(futureActiveBalance == 0 || futureActiveBalance >= minActiveBalance, ERROR_INVALID_DEACTIVATION_AMOUNT);
 
+        // Check future balance is not bellow the total deactivation locked of the guardian
```
sohkai:
```suggestion
        // Check future balance is not below the total activation lock of the guardian
```

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -164,6 +179,10 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
...
+        uint256 totalDeactivationLock = guardian.deactivationLocks.total;
```
sohkai: I was thinking about the concept, maybe we want to equate this to "minimum active balance" instead?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -669,6 +749,9 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
         // the activation amount since we have just added it to the available balance of the guardian.
         if (_data.toBytes4() == GuardiansRegistry(this).activate.selector) {
             _activateTokens(_guardian, _amount, _from);
+        } else if (_data.toBytes4() == GuardiansRegistry(this).lockDeactivation.selector) {
+            _activateTokens(_guardian, _amount, _from);
+            _lockDeactivation(_guardian, _from, _amount);
```
sohkai:
```suggestion
            _lockDeactivation(_guardian, msg.sender, _amount);
```

Would this be more correct?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -207,6 +226,37 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
...
+    * @param _guardian Address of the guardian unlocking the deactivation amount of
+    * @param _lockManager Address of the lock manager controlling the lock
+    * @param _amount Amount of tokens to be unlocked for deactivation
+    */
+    function unlockDeactivation(address _guardian, address _lockManager, uint256 _amount) external {
+        require(ILockManager(_lockManager).canUnlock(_guardian, _amount), ERROR_CANNOT_UNLOCK_DEACTIVATION);
+
+        DeactivationLocks storage deactivationLocks = guardiansByAddress[_guardian].deactivationLocks;
+        uint256 lockedAmount = deactivationLocks.lockedBy[_lockManager];
+        require(lockedAmount > 0, ERROR_SENDER_NOT_LOCK_MANAGER);
```
sohkai: Technically this error is more like "nothing is locked for you by your given lock manager," right?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -207,6 +226,37 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
...
+    }
+
+    /**
+    * @notice Unlock `@tokenAmount(self.token(), _amount)` from deactivation for `_guardian`
+    * @param _guardian Address of the guardian unlocking the deactivation amount of
+    * @param _lockManager Address of the lock manager controlling the lock
+    * @param _amount Amount of tokens to be unlocked for deactivation
+    */
+    function unlockDeactivation(address _guardian, address _lockManager, uint256 _amount) external {
+        require(ILockManager(_lockManager).canUnlock(_guardian, _amount), ERROR_CANNOT_UNLOCK_DEACTIVATION);
```
sohkai: We may want to check if the `msg.sender` is the lock manager and skip this check if so.

sohkai (commented):
Couple of initial notes!

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: Sounds better! :)

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: Note that `_from` is actually the `msg.sender` in `stake` and `stakeFor`, and we are trusting the token in `receiveApproval`. Although the last will change once we support permit

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: Isn't going to increase the gas usage? cause we will need to encode one extra word parameterizing a boolean to tell whether the lock has increased or not, while this way simply increases the bytecode size. Although it's worth testing it

packages/evm/test/helpers/utils/guardians.js
```diff
@@ -1,6 +1,7 @@
 const { sha3, toChecksumAddress } = require('web3-utils')
 
 const ACTIVATE_DATA = sha3('activate(uint256)').slice(0, 10)
+const LOCK_ACTIVATION_DATA = sha3('lockActivation(address,uint256)').slice(0, 10)
```
sohkai: Ah, I think I mentioned this in another PR, but do you think it'd be worth pulling out the ABI encoder for this?

docs/4-entry-points/4-guardians-registry.md
```diff
@@ -113,7 +113,38 @@ This module is in the one handling all the staking/unstaking logic for the guard
...
+    - Increase the amount locked for the guardian by the given lock manager
+
+### 4.4.9. Unlock activation
+
+- **Actor:** External entity incentivized to unlock the activation of a guardian of the Protocol
+- **Inputs:**
+    - **Guardian:** Address of the guardian unlocking the active balance of
+    - **Lock manager:** Address of the lock manager controlling the lock
+    - **Amount:** Amount of active tokens to be unlocked
+    - **Deactivate:** Whether the requested amount must be deactivated too
```
sohkai:
```suggestion
    - **Deactivate:** Whether the requested amount should be deactivated immediately
```

docs/4-entry-points/4-guardians-registry.md
```diff
@@ -221,7 +252,18 @@ This module is in the one handling all the staking/unstaking logic for the guard
 - **State transitions:**
     - Update the total active balance limit
 
-### 4.4.16. Recover funds
+### 4.4.18. Change lock manager
```
sohkai, line 255: I think this is a bit confusing as a title, as rather than changing a singular lock manager, we're updating a whitelist

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -103,20 +121,26 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
...
+    event GuardianActivationLockIncreased(address indexed guardian, address indexed lockManager, uint256 amount, uint256 total);
+    event GuardianActivationLockDecreased(address indexed guardian, address indexed lockManager, uint256 amount, uint256 total);
     event GuardianBalanceLocked(address indexed guardian, uint256 amount);
     event GuardianBalanceUnlocked(address indexed guardian, uint256 amount);
     event GuardianSlashed(address indexed guardian, uint256 amount, uint64 effectiveTermId);
     event GuardianTokensAssigned(address indexed guardian, uint256 amount);
     event GuardianTokensBurned(uint256 amount);
     event GuardianTokensCollected(address indexed guardian, uint256 amount, uint64 effectiveTermId);
     event TotalActiveBalanceLimitChanged(uint256 previousTotalActiveBalanceLimit, uint256 currentTotalActiveBalanceLimit);
+    event LockManagerChanged(address indexed lockManager, bool allowed);
```
sohkai:
```suggestion
    event LockManagerWhitelistChanged(address indexed lockManager, bool allowed);
```

Maybe this is clearer :)

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -103,20 +121,26 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
...
+
     // Tree to store guardians active balance by term for the drafting process
     HexSumTree.Tree internal tree;
 
     event GuardianActivated(address indexed guardian, uint64 fromTermId, uint256 amount, address sender);
     event GuardianDeactivationRequested(address indexed guardian, uint64 availableTermId, uint256 amount);
     event GuardianDeactivationProcessed(address indexed guardian, uint64 availableTermId, uint256 amount, uint64 processedTermId);
     event GuardianDeactivationUpdated(address indexed guardian, uint64 availableTermId, uint256 amount, uint64 updateTermId);
+    event GuardianActivationLockIncreased(address indexed guardian, address indexed lockManager, uint256 amount, uint256 total);
+    event GuardianActivationLockDecreased(address indexed guardian, address indexed lockManager, uint256 amount, uint256 total);
```
sohkai: Going off of the prior conversation, I was thinking we wouldn't even need to include whether it was an increase or decrease. The event just tells us it changed, and provides the new amounts (or event trigger) so we can update off-chain state accordingly.

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -402,6 +452,16 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
         _setTotalActiveBalanceLimit(_totalActiveBalanceLimit);
     }
 
+    /**
+    * @notice `_allowed ? 'Allow' : 'Disallow'` `_lockManager` as a lock manager
+    * @param _lockManager Address of the lock manager to be changed
+    * @param _allowed Whether the lock manager is allowed
+    */
+    function changeLockManager(address _lockManager, bool _allowed) external onlyConfigGovernor {
```
sohkai: Related to above, maybe we could rename to `updateLockManagerWhitelist()` or `changeLockManagerStatus()`?

I'm also wondering the tradeoff between keeping this configuration here, vs. in the Protocol config. I assume the ability to persist this whitelist over individual registry upgrades wouldn't be worth the extra gas cost and complexity?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -536,6 +609,15 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
         return totalActiveBalanceLimit;
     }
 
+    /**
+    * @dev Tell whether a lock manager is allowed
+    * @param _lockManager Address of the lock manager being queried
+    * @return True if the lock manager is allowed
+    */
+    function isLockManagerAllowed(address _lockManager) external view returns (bool) {
```
sohkai: Maybe `isLockManagerWhitelisted()` is more appropriate if we make the above changes

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -654,6 +762,25 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
...
+        bool isAllowed = _isLockManagerAllowed(_lockManager) || _isLockManagerAllowed(ANY_ENTITY);
+        require(isAllowed, ERROR_LOCK_MANAGER_NOT_ALLOWED);
+
+        ActivationLocks storage activationLocks = guardiansByAddress[_guardian].activationLocks;
+        uint256 newTotalLocked = activationLocks.total.add(_amount);
+        uint256 newLockedAmount = activationLocks.lockedBy[_lockManager].add(_amount);
+
+        activationLocks.total = newTotalLocked;
+        activationLocks.lockedBy[_lockManager] = newLockedAmount;
+        emit GuardianActivationLockIncreased(_guardian, _lockManager, newLockedAmount, newTotalLocked);
```
sohkai: Just want to sanity check our expected interaction flow, since this doesn't call back to the `lockManager`.

Creating a lock could happen in two ways:

- You call the lock manager, they stake and lock for you
  - Any internal accounting that needs to be updated in the lock manager can be transparently done here, as it knows about your request
- You call the registry and specify the lock manager there
  - Someone needs to manually trigger the lock manager afterwards to update its accounting, based on a call to `getActivationLock()`
  - If you never do this, you should be able to recover if the lock manager is properly implemented; it should allow withdrawals of those who aren't in its internal accounting

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -207,6 +219,44 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
...
+    * @param _deactivate Whether the requested amount must be deactivated too
+    */
+    function unlockActivation(address _guardian, address _lockManager, uint256 _amount, bool _deactivate) external {
+        bool canUnlock = _lockManager == msg.sender || ILockManager(_lockManager).canUnlock(_guardian, _amount);
+        require(canUnlock, ERROR_CANNOT_UNLOCK_ACTIVATION);
+
+        ActivationLocks storage activationLocks = guardiansByAddress[_guardian].activationLocks;
+        uint256 lockedAmount = activationLocks.lockedBy[_lockManager];
+        require(lockedAmount > 0, ERROR_ZERO_LOCK_ACTIVATION);
+        require(_amount > 0, ERROR_UNLOCK_ACTIVATION_ZERO);
```
sohkai: I wonder if we should change this so it acts more like the deactivation?

If we do the `canUnlock()` check after this, could we have `amount = lockedAmount` if given 0?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -207,6 +219,44 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, ERC900,
...
+
+        uint256 unlockedAmount = lockedAmount > _amount ? _amount : lockedAmount;
+        uint256 newLockedAmount = lockedAmount.sub(unlockedAmount);
+        uint256 newTotalLocked = activationLocks.total.sub(unlockedAmount);
+
+        activationLocks.total = newTotalLocked;
+        activationLocks.lockedBy[_lockManager] = newLockedAmount;
+        emit GuardianActivationLockDecreased(_guardian, _lockManager, newLockedAmount, newTotalLocked);
+
+        if (msg.sender == _guardian && _deactivate) {
```
sohkai, line 254: Hmm, would it be a problem if the lock manager wanted to do this?

packages/evm/test/registry/guardians-registry-locking.js
```diff
@@ -0,0 +1,265 @@
...
+      allowLockManager()
+
+      context('when any lock manager is allowed', async () => {
+        allowLockManager(true)
+        itCreatesTheActivationLock()
+      })
+
+      context('when any lock manager is not allowed', async () => {
+        itCreatesTheActivationLock()
+      })
```
sohkai, line 108: Might be worth adding a quick note explaining the situation in the two cases above; was initially a bit confused on the additional use of (or lack of) `allowLockManager()`.

packages/evm/test/registry/guardians-registry-locking.js
```diff
@@ -0,0 +1,265 @@
...
+
+          context('when the sender is not the guardian', () => {
+            const from = someone
+
+            itUnlocksTheActivation(from)
+
+            it('can request a deactivation in the same call', async () => {
+              await activateTokens(lockAmount)
+
+              const receipt = await registry.unlockActivation(guardian, lockManager.address, unlockAmount, true, { from })
```
sohkai, line 196: Shouldn't this fail?

packages/evm/test/registry/guardians-registry-staking.js
```diff
@@ -702,6 +702,223 @@ contract('GuardiansRegistry', ([_, guardian, anotherGuardian]) => {
...
+          const termId = await controller.getLastEnsuredTermId()
+          const currentTermPreviousBalance = await registry.activeBalanceOfAt(recipient, termId)
+
+          await registry.stakeFor(recipient, amount, data, { from })
+
+          const currentTermCurrentBalance = await registry.activeBalanceOfAt(recipient, termId)
+          assertBn(currentTermPreviousBalance, currentTermCurrentBalance, 'current term active balances do not match')
+        })
+
+        if (recipient !== from) {
```
sohkai, line 746: Should we also add a test for the locked amounts? The recipient should be the one gaining a lock but not the sender.

packages/evm/test/registry/guardians-registry-staking.js
```diff
@@ -702,6 +702,223 @@ contract('GuardiansRegistry', ([_, guardian, anotherGuardian]) => {
...
+        await registry.changeLockManager(guardian, true, { from: governor })
+      })
+
+      const itHandlesStakesWithLockActivationProperlyFor = (recipient, amount, data) => {
+        it('creates the lock', async () => {
+          await registry.stakeFor(recipient, amount, data, { from })
+
+          const { amount: lockedAmount, total } = await registry.getActivationLock(recipient, guardian)
+          assertBn(lockedAmount, amount, 'locked amount does not match')
+          assertBn(total, amount, 'total locked amount does not match')
```
sohkai, line 720: We should also check the lock manager

sohkai (commented):
Couple of additional thoughts past the last review. Most pressing are those related to the locking/unlocking flow.

packages/evm/test/registry/guardians-registry-staking.js
facuspagnuolo: Note that I'm using the guardian as the lock manager, is part of the query

packages/evm/test/registry/guardians-registry-locking.js
facuspagnuolo: what's the reasoning?

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: In that case the lock manager couldn't simply answer `true` to `canUnlock` so the guardian has complete control on his tokens?

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: > I'm also wondering the tradeoff between keeping this configuration here, vs. in the Protocol config. I assume the ability to persist this whitelist over individual registry upgrades wouldn't be worth the extra gas cost and complexity?

I'd benefit the guardians instead of the governor here tbh, I put it here cause it wasn't shared between other modules

packages/evm/test/registry/guardians-registry-locking.js
facuspagnuolo: Discussed offline: it's fine cause anyone can unlock activations

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: agreed offline to avoid having the lock managers allowed to deactivate tokens

sohkai (commented):
All good for me!

PR #16, opened 2020-10-24
Registry: Implement activators whitelist

facuspagnuolo:


sohkai:
More of a meta-note, but perhaps it'd be useful to pull the first two commits out to a separate PR. I don't think they conflict with anything afterwards, and it'd be good to move them as they're addressing previous feedback.

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -140,6 +144,7 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
     event GuardianTokensBurned(uint256 amount);
     event GuardianTokensCollected(address indexed guardian, uint256 amount, uint64 effectiveTermId);
     event TotalActiveBalanceLimitChanged(uint256 previousTotalActiveBalanceLimit, uint256 currentTotalActiveBalanceLimit);
+    event ActivatorChanged(address indexed activator, bool allowed);
```
sohkai: Similar to #15, I think this event's name is a bit ambiguous. Perhaps `ActivatorWhitelistChanged` would be clearer?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -609,6 +624,15 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
         return totalActiveBalanceLimit;
     }
 
+    /**
+    * @dev Tell whether a activator is allowed
```
sohkai:
```suggestion
    * @dev Tell whether an activator is allowed
```

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -636,6 +660,10 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
     function _activateTokens(address _guardian, uint256 _amount, address _sender) internal {
         uint64 termId = _ensureCurrentTerm();
 
+        // Make sure the sender is allowed to activate tokens on behalf of the guardian
+        bool isAllowed = _guardian == _sender || _isActivatorAllowed(_sender) || _isActivatorAllowed(ANY_ENTITY);
```
sohkai: ⚠️ I'm having second thoughts about allowing `ANY_ENTITY` here.

If we allow it here, anyone can call `stakeFor()` for anyone with an existing allowance.

-----

For locking, on further thought, I don't think `ANY_ENTITY` presents any immediate threat outside of a user misconfiguring their lock manager to an incorrect address and thereby being locked out.

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -769,7 +797,7 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
     * @param _amount Amount of tokens to be added to the activation locked amount of the guardian
     */
     function _lockActivation(address _guardian, address _lockManager, uint256 _amount) internal {
-        bool isAllowed = _isLockManagerAllowed(_lockManager) || _isLockManagerAllowed(ANY_ENTITY);
+        bool isAllowed = _isLockManagerAllowed(_lockManager) || _isLockManagerAllowed(ANY_ENTITY) || _guardian == _lockManager;
```
sohkai: Not sure I understand this change; would we ever expect a guardian to use themselves as a lock manager (even if the guardian was a contract)?

packages/evm/test/registry/guardians-registry-staking.js
```diff
@@ -677,13 +677,43 @@ contract('GuardiansRegistry', ([_, guardian, anotherGuardian, governor]) => {
...
+              const receipt = await registry.changeActivator(from, true, { from: governor })
+
+              assertAmountOfEvents(receipt, REGISTRY_EVENTS.ACTIVATOR_CHANGED)
+              assertEvent(receipt, REGISTRY_EVENTS.ACTIVATOR_CHANGED, { expectedArgs: { activator: from, allowed: true } })
+            })
+
+            itHandlesStakesWithActivationProperlyForDifferentAmounts(recipient, data)
+          })
+
+          context('when the sender is allowed as activator', () => {
```
sohkai:
```suggestion
          context('when the sender is not allowed as activator', () => {
```

packages/evm/test/registry/guardians-registry-staking.js
```diff
@@ -677,13 +677,43 @@ contract('GuardiansRegistry', ([_, guardian, anotherGuardian, governor]) => {
...
```
sohkai:
```suggestion
          context('when the sender is not allowed as activator', () => {
```

packages/evm/test/registry/guardians-registry-staking.js
```diff
@@ -894,13 +924,43 @@ contract('GuardiansRegistry', ([_, guardian, anotherGuardian, governor]) => {
...
+              const receipt = await registry.changeActivator(from, true, { from: governor })
+
+              assertAmountOfEvents(receipt, REGISTRY_EVENTS.ACTIVATOR_CHANGED)
+              assertEvent(receipt, REGISTRY_EVENTS.ACTIVATOR_CHANGED, { expectedArgs: { activator: from, allowed: true } })
+            })
+
+            itHandlesStakesWithLockActivationProperlyForDifferentAmounts(recipient, data)
+          })
+
+          context('when the sender is allowed as activator', () => {
```
sohkai:
```suggestion
          context('when the sender is not allowed as activator', () => {
```

packages/evm/test/registry/guardians-registry-staking.js
```diff
@@ -894,13 +924,43 @@ contract('GuardiansRegistry', ([_, guardian, anotherGuardian, governor]) => {
...
```
sohkai:
```suggestion
          context('when the sender is not allowed as activator', () => {
```

packages/evm/test/registry/guardians-registry-staking.js
```diff
@@ -459,7 +459,7 @@ contract('GuardiansRegistry', ([_, guardian, anotherGuardian, governor]) => {
       })
```
sohkai, line 459: I didn't notice any tests for ANY_ENTITY activators, which may be another sign that we may want to remove it :)

sohkai (commented):
Bunch of suggestions on fixing test names that were copied, but otherwise, I had two concerns with the logic changes introduced for activating and locking.

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: Rolling it back

sohkai (approved):
LGTM!

PR #18, opened 2020-10-26
Contracts: Reorganize files

facuspagnuolo:


sohkai (approved):
❤️ This looks much, much better!

PR #19, opened 2020-10-26
Contracts: Architecture enhancements

facuspagnuolo:


packages/evm/contracts/AragonProtocol.sol
```diff
@@ -93,20 +92,37 @@ contract AragonProtocol is Controller, IArbitrator {
     * @param _evidence Data submitted for the evidence related to the dispute
     */
     function submitEvidence(uint256 _disputeId, address _submitter, bytes calldata _evidence) external {
```
sohkai, line 94: This is a nitpick (we can lift as an issue), we should update the entrypoint docs to include these new interfaces :)

packages/evm/contracts/AragonProtocol.sol
```diff
@@ -93,20 +92,37 @@ contract AragonProtocol is Controller, IArbitrator {
...
+
+    /**
+    * @notice Submit `_evidence` as evidence from `_submitter` for dispute #`_disputeId`
+    * @dev This entry point can be used to submit evidences for previous Dispute Manager instances
+    * @param _disputeManager Dispute manager to be used
+    * @param _disputeId Id of the dispute in the Protocol
+    * @param _submitter Address of the account submitting the evidence
+    * @param _evidence Data submitted for the evidence related to the dispute
+    */
+    function submitEvidenceFor(IDisputeManager _disputeManager, uint256 _disputeId, address _submitter, bytes calldata _evidence) external {
```
sohkai, line 106: I'm not against this naming, but I think it could be a bit confusing given how we think about `stakeFor()` (and `activateFor()`, etc).

Maybe `ForModule()` or `OnSpecificModule()` or etc. would be clearer?

packages/evm/contracts/AragonProtocol.sol
```diff
@@ -93,20 +92,37 @@ contract AragonProtocol is Controller, IArbitrator {
...
+        _submitEvidence(_disputeManager, _disputeId, _submitter, _evidence);
```
sohkai, line 107: ⚠️ Wondering if we should double check that the given module is in the historical list?

sohkai (commented):
Really like these changes, they look really good!

facuspagnuolo:
Addressed these comments in https://github.com/aragon/protocol/pull/34

PR #20, opened 2020-10-26
Chore: Multiple enhancements

facuspagnuolo:


packages/evm/test/core/controller-modules.js
```diff
@@ -604,11 +604,11 @@ contract('Controller', ([_, fundsGovernor, configGovernor, modulesGovernor, some
       const from = someone
 
       it('reverts', async () => {
-        await assertRevert(controller.linkModules([ZERO_ADDRESS], [ZERO_BYTES32], { from }), CONTROLLER_ERRORS.SENDER_NOT_GOVERNOR)
+        await assertRevert(controller.syncModuleLinks([ZERO_ADDRESS], [ZERO_BYTES32], { from }), CONTROLLER_ERRORS.SENDER_NOT_GOVERNOR)
       })
     })
 
-    context('when trying to call it directly', () => {
+    context('when trying to call linkModules directly', () => {
```
sohkai, line 611: Just wondering, should we move this test to the `Controlled` tests instead?

packages/evm/contracts/core/modules/Controlled.sol
```diff
@@ -25,7 +25,6 @@ contract Controlled is IModulesLinker, IsContract, ModuleIds, ConfigConsumer {
     string private constant ERROR_SENDER_NOT_ACTIVE_DISPUTE_MANAGER = "CTD_SEND_NOT_ACTIVE_DISPUTE_MGR";
     string private constant ERROR_SENDER_NOT_CURRENT_DISPUTE_MANAGER = "CTD_SEND_NOT_CURRENT_DISPUTE_MGR";
 
-    // Note: I noticed that we usually don't leave storage public, but add different getters with a `get*` prefix?
```
sohkai: Actually I realized we do have some public state variables and only use the `get*` prefix for more complicated getters.

packages/evm/test/core/controller-modules.js
facuspagnuolo: I think it's fine to keep it here, I consider it a particular case of the functionality being tested

PR #24, opened 2020-10-29
Contracts: Model signatures validator

facuspagnuolo:
This PR aims to provide a new signature validation module that can be extended to other modules to support "meta-txs"-like methods so users can authorize other accounts with a signed message to operate on their behalf


facuspagnuolo:
Closing in favor of #25, #26, and #27

PR #25, opened 2020-10-30
Modules: Support whitelisted relayers

facuspagnuolo:


packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
+pragma solidity ^0.5.17;
```
sohkai: More of a meta discussion, but wondering if this contract is better as a `/lib` contract.

Does it make more sense being in `/core` because it includes a storage variable?

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+
+import "../../lib/utils/TimeHelpers.sol";
+import "../../lib/standards/IERC712.sol";
+
+
+contract SignaturesValidator is IERC712, TimeHelpers {
+    string private constant ERROR_INVALID_SIGNATURE = "SV_INVALID_SIGNATURE";
+
+    // [v,r,s] signature + deadline
+    uint256 internal constant EXTRA_CALLDATA_LENGTH = 32 * 4;
```
sohkai: Depending on how we'd like to encode the signature, usually `v` is uint8 so we could save _just_ a little bit of calldata if we made that assumption.

I can see how doing this could be a bit of a nightmare in assembly though, and I'm happy to let it go :).

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+    bytes32 internal constant NAME_HASH = 0xd29d26249bb0c8fe08bcf70d00b6f5b6b54b653b5a7e4157b490095bbb233349;
+    // bytes32 private constant VERSION_HASH = keccak256("1")
+    bytes32 internal constant VERSION_HASH = 0xc89efdaa54c0f20c7adf612882df0950f5a951637e0307cdcb4c672f298b8bc6;
+
+    modifier authenticate(address _user) {
+        _validateSignature(_user);
+        _;
+    }
+
+    mapping (address => uint256) internal nextNonce;
```
sohkai:
```suggestion
    mapping (address => uint256) public nonces;
```

I know we usually use the `get()` prefix, but perhaps we can stick to `nonces` since it's pseudo-standardized with ERC-2612 :).

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+
+    constructor () public {
```
sohkai: Are we able to remove this?

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+    // [v,r,s] signature + deadline
+    uint256 internal constant EXTRA_CALLDATA_LENGTH = 32 * 4;
+    // bytes32 private constant EIP712DOMAIN_HASH = keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)")
+    bytes32 internal constant EIP712DOMAIN_HASH = 0x8b73c3c69bb8fe3d512ecc4cf759cc79239f7b179b0ffacaa9a75d522b39400f;
+    // bytes32 private constant NAME_HASH = keccak256("Aragon Protocol")
+    bytes32 internal constant NAME_HASH = 0xd29d26249bb0c8fe08bcf70d00b6f5b6b54b653b5a7e4157b490095bbb233349;
+    // bytes32 private constant VERSION_HASH = keccak256("1")
+    bytes32 internal constant VERSION_HASH = 0xc89efdaa54c0f20c7adf612882df0950f5a951637e0307cdcb4c672f298b8bc6;
+
+    modifier authenticate(address _user) {
```
sohkai:
```suggestion
    modifier authenticateSender(address _user) {
```

Perhaps this is a bit more clear, since `authenticate()` can be a bit ambiguous.

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+    * @dev Validate signature
+    */
+    function _validateSignature(address _user) internal {
+        require(_isSignatureValid(_user), ERROR_INVALID_SIGNATURE);
+    }
+
+    /**
+    * @dev Tell whether a signature is valid and update nonce
+    */
+    function _isSignatureValid(address _user) internal returns (bool) {
```
sohkai: If we don't have a plan to use both overloads, I'd prefer to just stick to this one :).

packages/evm/contracts/lib/standards/IERC712.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.5.17;
```
sohkai: Haha what do you think about removing this file? This is technically not quite an ERC-interface (no interface to get the domain separator is specified in the ERC itself, and ERC-2612 is now proposing `DOMAIN_SEPARATOR()`).

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+    /**
+    * @dev Decode word from extra calldata
+    */
+    function _decodeExtraCalldataWord(uint256 _offset) internal pure returns (uint256 result) {
+        uint256 offset = _offset;
+        assembly {
+            let ptr := mload(0x40)
+            mstore(0x40, add(ptr, 0x20))
+            calldatacopy(ptr, sub(calldatasize, sub(EXTRA_CALLDATA_LENGTH, offset)), 0x20)
+            result := mload(ptr)
```
sohkai: Sanity checking, could we use a `calldataload` here instead?

Also note that there's no explicit check on `calldatasize` here, which may result in garbage being loaded into the ptr in a way that's not possible with `_calldata()`. It's not a big deal (I think accessing out of bounds calldata doesn't revert?), but we likely want align on whether or not to protect for this case.

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+    * @dev Tell whether a signature is valid
+    */
+    function _isSignatureValid(address _user, uint256 _nonce) internal view returns (bool) {
+        if (msg.sender == _user) {
+            return true;
+        }
+
+        uint256 deadline = _deadline();
+        (uint8 v, bytes32 r, bytes32 s) = _signature();
+        bytes32 encodeData = keccak256(abi.encode(_calldata(), msg.sender, _nonce, deadline));
```
sohkai: Ah, in order for this to be ERC-712 compliant, we need to include a typehash here.

I think we were discussing how we would pass this typehash down, but it could be specified as an argument in the modifier.

Theoretically all the calldata should be ABI-encoded already, so passing down a typehash like `keccak256("Stake(uint256 amount,address spender,uint256 nonce,uint256 deadline)")` and concating the calldata and remaining signature arguments (`sender, nonce, deadline`) together could work.

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+        s = bytes32(_decodeExtraCalldataWord(0x60));
+    }
+
+    /**
+    * @dev Decode original calldata
+    */
+    function _calldata() internal pure returns (bytes memory result) {
+        result = msg.data;
+        if (result.length > EXTRA_CALLDATA_LENGTH) {
+            assembly { mstore(result, sub(calldatasize, EXTRA_CALLDATA_LENGTH)) }
```
sohkai: Discussed offline, but just leaving a note so we don't forget:

Perhaps we can reformulate this in a way where we copy only the parts of `msg.data` that we care about.

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+    function _signature() internal pure returns (uint8 v, bytes32 r, bytes32 s) {
+        v = uint8(_decodeExtraCalldataWord(0x20));
+        r = bytes32(_decodeExtraCalldataWord(0x40));
+        s = bytes32(_decodeExtraCalldataWord(0x60));
+    }
+
+    /**
+    * @dev Decode original calldata
+    */
+    function _calldata() internal pure returns (bytes memory result) {
```
sohkai: Since this is only used internally, perhaps we can make this `private` and rename it to something like `_calldataWithoutSignature()`

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        uint256 deadline = _deadline();
+        (uint8 v, bytes32 r, bytes32 s) = _signature();
+        bytes32 encodeData = keccak256(abi.encode(_calldata(), msg.sender, _nonce, deadline));
+        bytes32 digest = keccak256(abi.encodePacked("\x19\x01", _getDomainSeparator(), encodeData));
+
+        address recoveredAddress = ecrecover(digest, v, r, s);
+        // Explicitly disallow authorizations for address(0) as ecrecover returns address(0) on malformed messages
+        return deadline >= getTimestamp() && recoveredAddress != address(0) && recoveredAddress == _user;
```
sohkai: What do you think about using a different require message if the deadline has passed?

packages/evm/test/core/signature-validator.js
```diff
@@ -0,0 +1,611 @@
...
+const { keccak256, toHex, padLeft, soliditySha3 } = require('web3-utils')
+const { assertAmountOfEvents, assertEvent, assertRevert, assertBn } = require('@aragon/contract-helpers-test/src/asserts')
+
+const { SIGNATURES_VALIDATOR_ERRORS } = require('../helpers/utils/errors')
+
+const SignaturesValidator = artifacts.require('SignaturesValidatorMock')
+
+contract('SignaturesValidator', ([_, sender, strange]) => {
+  let validator
+  const wallet = web3.eth.accounts.create('erc3009')
```
sohkai: Hehe, let's use a different string :)

packages/evm/test/core/signature-validator.js
```diff
@@ -0,0 +1,611 @@
...
+    context('when there is no signature encoded', () => {
+      it('decodes empty data', async () => {
+        const receipt = await validator.sendTransaction({ data: calldata, from: sender })
+
+        assertAmountOfEvents(receipt, 'CalldataDecoded')
+        assertEvent(receipt, 'CalldataDecoded', { expectedArgs: { data: calldata, deadline: 0, v: 0, r: ZERO_BYTES32, s: ZERO_BYTES32 } })
+      })
+    })
+
+    context('when there is a signature encoded', () => {
```
sohkai: As an additional test, perhaps it's worth calling a function with multiple arguments (and a dynamic-length argument) :).

packages/evm/test/core/signature-validator.js
```diff
@@ -0,0 +1,611 @@
...
+
+  describe('authenticate', () => {
+    let user, deadline, nonce, allowedSender, allowedFunctionality
+
+    before('mock nonces', async () => {
+      await validator.increaseNonce(externalAccount)
+      await validator.increaseNonce(sender)
+    })
+
+    const itReverts = (extraCalldata = undefined) => {
```
sohkai: Nitpick: `undefined` is already the default

packages/evm/test/core/signature-validator.js
```diff
@@ -0,0 +1,611 @@
...
+      beforeEach(`set deadline with offset ${offset}`, async () => {
+        const currentTimestamp = await validator.getTimestampExt()
+        deadline = currentTimestamp.add(bn(offset))
+      })
+    }
+
+    context('when there is no extra calldata given', () => {
+      const extraCalldata = ''
+
+      context('when there sender and the user are the same', () => {
```
sohkai:
```suggestion
      context('when the sender and the user are the same', () => {
```

I saw a couple other repetitions of this, probably an easy search and replace :)

packages/evm/test/core/signature-validator.js
```diff
@@ -0,0 +1,611 @@
...
+        setUser(sender)
```
sohkai: Perhaps it'd be useful to explicitly call `setAuthorizedSender()` to make this clear :)

packages/evm/test/core/signature-validator.js
```diff
@@ -0,0 +1,611 @@
...
+      context('when the sender and the user are the same', () => {
+        setUser(sender)
+
+        context('when the extra calldata is malformed', () => {
+          const extraCalldata = 'abcd'
+
+          itAllowsTheSender(extraCalldata)
+        })
+
+        context('when the extra calldata is well formed', () => {
```
sohkai: Just clarifying, here when we set bad variables (e.g. `setDeadline(-100)`), it doesn't matter because the extra calldata isn't checked?

sohkai (commented):
I have a couple of questions and notes related to the core validation design, and as discussed offline, I'd like to compare this against an off-protocol wrapper.

The tests look really good for this though!

packages/evm/contracts/core/modules/SignaturesValidator.sol
facuspagnuolo: actually `v` is only a byte, let's optimize it

packages/evm/contracts/core/modules/SignaturesValidator.sol
facuspagnuolo: Let me know what's your opinion about this based on the latest change I introduced for this logic

packages/evm/contracts/core/modules/SignaturesValidator.sol
facuspagnuolo: This one is actually being used, especially in those cases where we the sender could be someone allowed by config (e.g. whitelisted lock managers, activators, or representatives)

packages/evm/contracts/core/modules/SignaturesValidator.sol
facuspagnuolo: I changed the way we decode the calldata, let me know your thoughts

packages/evm/contracts/core/modules/SignaturesValidator.sol
facuspagnuolo: This one is no longer available

packages/evm/test/core/signature-validator.js
facuspagnuolo: I changed these in a further PR (see d7e01f6be6c9eaeab8b117e4bff5e64db07a46ce)

packages/evm/test/core/signature-validator.js
facuspagnuolo: Yup, but I wanted to make explicit it actually has a purpose if you pass an undefined extra calldata, do you see it useless?

packages/evm/test/core/signature-validator.js
facuspagnuolo: The `setDeadline` argument is actually an offset for the current timestamp

packages/evm/contracts/test/core/modules/SignaturesValdiatorMock.sol
```diff
@@ -0,0 +1,29 @@
+pragma solidity ^0.5.17;
```
izqui: Small typo in file name

packages/evm/contracts/core/modules/SignaturesValidator.sol
```diff
@@ -84,41 +83,20 @@ contract SignaturesValidator is IERC712, TimeHelpers {
...
+    function _decodeCalldata() internal pure returns (bytes memory data, uint256 deadline, bytes32 r, bytes32 s, uint8 v) {
+        data = msg.data;
+        if (data.length > EXTRA_CALLDATA_LENGTH) {
+            assembly {
+                let realCalldataSize := sub(calldatasize, EXTRA_CALLDATA_LENGTH)
+                let extraCalldataPtr := add(add(data, 0x20), realCalldataSize)
+                deadline := mload(extraCalldataPtr)
+                r := mload(add(extraCalldataPtr, 0x20))
+                s := mload(add(extraCalldataPtr, 0x40))
+                v := byte(0, mload(add(extraCalldataPtr, 0x60)))
```
sohkai: Oh nice, never knew about the `byte()` helper in assembly!

Wondering if it'd be too sketchy to put `v` first, since that can be quite common and the ABI-encoding of this would match `uint8 v, bytes32 r, bytes32 s` found elsewhere. We would then be offsetting the rest by `0x21` and `0x41`?

packages/evm/test/core/signature-validator.js
sohkai: Got it, perhaps a comment would suffice :). I don't know if this communicates that well, since I'm already thinking "well `extraCalldata` would be undefined anyway by default".

packages/evm/test/core/signature-validator.js
sohkai: Right, that's why I wasn't sure. Using `setDeadline(-100)` (and likewise, other bad constructions in this context) should fail, but they're expected to pass.

packages/evm/contracts/core/modules/ControlledRelayable.sol
```diff
@@ -0,0 +1,37 @@
...
+
+import "./Controlled.sol";
+
+
+contract ControlledRelayable is Controlled {
+    string private constant ERROR_SENDER_NOT_ALLOWED = "CTD_SENDER_NOT_ALLOWED";
+
+    /**
+    * @dev Modifier for modules to support relayed transactions.
+    *      This modifier will check that the sender is the user acting on behalf of or a whitelisted relayer.
```
sohkai:
```suggestion
    *      This modifier will check that the sender is the user to act on behalf of or a whitelisted relayer.
```

packages/evm/contracts/core/modules/ControlledRelayable.sol
```diff
@@ -0,0 +1,37 @@
...
+    * @param _user Address of the user executing an action for
```
sohkai:
```suggestion
    * @param _user Address of the user to act on behalf of
```

packages/evm/contracts/core/modules/ControlledRelayable.sol
```diff
@@ -0,0 +1,37 @@
...
+    /**
+    * @dev Constructor function
+    * @param _controller Address of the controller
+    */
+    constructor(Controller _controller) Controlled(_controller) public {
+        // solium-disable-previous-line no-empty-blocks
+    }
+
+    /**
+    * @dev Tell whether a sender is allowed or not
```
sohkai:
```suggestion
    * @dev Tell whether the sender is the user to act on behalf of or a whitelisted relayer
```

sohkai (approved):
Just a couple of tiny comment nitpicks. LGTM!

PR #26, opened 2020-10-30
Registry: Adopt new signatures verification flow

facuspagnuolo:


docs/4-entry-points/4-guardians-registry.md
```diff
@@ -20,131 +20,136 @@ This module is in the one handling all the staking/unstaking logic for the guard
     - Save the guardian token address
     - Save the total active balance limit
 
-### 4.4.2. Activate
+### 4.4.1. Stake
```
sohkai: Not sure, but `4.4.1` is taken by the constructor already, so should this be `4.4.2`?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -131,6 +130,8 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
     // Tree to store guardians active balance by term for the drafting process
     HexSumTree.Tree internal tree;
 
+    event Staked(address indexed user, uint256 amount, uint256 total, bytes data);
+    event Unstaked(address indexed user, uint256 amount, uint256 total, bytes data);
```
sohkai: Given that we've made the decision to move away from ERC-900 now, what do you think about dropping the `data` at the end? That way we'll also be able to simplify all the related functions that leave this param unused :).

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -168,78 +169,82 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
...
-    * @notice Activate `_amount == 0 ? 'all available tokens' : @tokenAmount(self.token(), _amount)` for the next term
-    * @param _amount Amount of guardian tokens to be activated for the next term
+    * @notice Stake `@tokenAmount(self.token(), _amount)` for `_guardian`
+    * @param _guardian Address of the guardian to stake tokens to
+    * @param _amount Amount of tokens to be staked
+    * @param _data Optional data is never used by this function, only logged
     */
-    function activate(uint256 _amount) external {
-        _activateTokens(msg.sender, _amount, msg.sender);
+    function stake(address _guardian, uint256 _amount, bytes calldata _data) external {
```
sohkai: Should this include an `authenticate(_guardian)`?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -810,26 +800,16 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
...
-        if (_data.toBytes4() == GuardiansRegistry(this).activate.selector) {
-            _activateTokens(_guardian, _amount, _from);
-        } else if (_data.toBytes4() == GuardiansRegistry(this).lockActivation.selector) {
-            _activateTokens(_guardian, _amount, _from);
-            _lockActivation(_guardian, _from, _amount);
-        }
-
         emit Staked(_guardian, _amount, _totalStakedFor(_guardian), _data);
-        require(guardiansToken.safeTransferFrom(_from, address(this), _amount), ERROR_TOKEN_TRANSFER_FAILED);
+        require(guardiansToken.safeTransferFrom(msg.sender, address(this), _amount), ERROR_TOKEN_TRANSFER_FAILED);
```
sohkai, line 801: ⚠️ shouldn't this be `_guardian`?

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -689,15 +682,15 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
         }
 
         _updateAvailableBalanceOf(_guardian, amountToActivate, false);
-        emit GuardianActivated(_guardian, nextTermId, amountToActivate, _sender);
+        emit GuardianActivated(_guardian, nextTermId, amountToActivate, msg.sender);
```
sohkai: Would agree that this last `sender` event param seems pretty unnecessary.

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -168,78 +169,82 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
...
     * @param _lockManager Address of the lock manager that will control the lock
     * @param _amount Amount of active tokens to be locked
     */
-    function lockActivation(address _lockManager, uint256 _amount) external {
-        _lockActivation(msg.sender, _lockManager, _amount);
+    function lockActivation(address _guardian, address _lockManager, uint256 _amount) external {
+        // Make sure the sender is the guardian, someone allowed by the guardian, or the lock manager itself
+        bool isLockManagerAllowed = msg.sender == _lockManager || _isSignatureValid(_guardian);
+        // Make sure that the given lock manager is whitelisted or that any entity actually is
+        bool isLockManagerWhitelisted = _isLockManagerWhitelisted(_lockManager) || _isLockManagerWhitelisted(ANY_ENTITY);
```
sohkai: ⚠️ Given this change, the `ANY_ENTITY` is pretty useless :). It would mean any entity would be allowed to select any guardian and lock an arbitrary amount of tokens.

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -168,78 +169,82 @@ contract GuardiansRegistry is ControlledRecoverable, IGuardiansRegistry, IERC900
...
-    * @param _amount Amount of tokens to be unstaked
+    * @notice Stake and activate `@tokenAmount(self.token(), _amount)` for `_guardian`
+    * @param _guardian Address of the guardian staking and activating tokens for
+    * @param _amount Amount of tokens to be staked and activated
     * @param _data Optional data is never used by this function, only logged
     */
-    function unstake(uint256 _amount, bytes calldata _data) external {
-        _unstake(msg.sender, _amount, _data);
-    }
+    function stakeAndActivate(address _guardian, uint256 _amount, bytes calldata _data) external {
```
sohkai: I'm having a hard time understanding the `stakeAndActivate()` flow, and how we expect to support lock managers with it. Let's discuss!

sohkai (commented):
A couple of initial notes, let's discuss!

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: Stake is the only one that is not harmful, and I think it should be fine we allow other people to stake on behalf of some guardians. It would be like signaling, they would be basically gifting money

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: As explained above, these assets should be transferred from the real sender in `stake`

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: Totally

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: Yea, I agree with that, let's remove this one

packages/evm/contracts/registry/GuardiansRegistry.sol
sohkai: Oh this is smart. This clarifies how `stakeAndActivate()` work as well for me then :).

sohkai (approved):
LGTM!

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -11,11 +11,11 @@ import "../lib/standards/IERC20.sol";
 import "./ILockManager.sol";
 import "./IGuardiansRegistry.sol";
 import "../core/modules/Controller.sol";
-import "../core/modules/SignaturesValidator.sol";
+import "../core/modules/ControlledRelayable.sol";
```
sohkai: Nitpick: `ControlledRecoverable` is higher, alphabetically :)

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -255,7 +251,7 @@ contract GuardiansRegistry is IGuardiansRegistry, ControlledRecoverable, Signatu
         emit GuardianActivationLockChanged(_guardian, _lockManager, newLockedAmount, newTotalLocked);
 
         if (_requestDeactivation) {
-            _validateSignature(_guardian);
+            _authenticateSender(_guardian);
```
sohkai, line 255: Let's leave a comment here about our expected flow (that for the deactivation to happen with the unlock, the request must originate from the owner).

packages/evm/test/registry/guardians-registry-locking.js
```diff
@@ -32,32 +27,22 @@ contract('GuardiansRegistry', ([_, guardian, governor]) => {
...
+    return registry.stakeAndActivate(guardian, amount, { from: guardian })
   }
 
-  const activate = async (recipient, amount, sender = guardian) => {
-    await ANT.generateTokens(sender, amount)
-    await ANT.approve(registry.address, amount, { from: sender })
-    let calldata = registry.contract.methods.stakeAndActivate(recipient, amount.toString()).encodeABI()
-    calldata = await encodeAuthorization(registry, recipient, externalAccountPK, calldata, sender)
-    return registry.sendTransaction({ from: sender, data: calldata })
+  const lockActivation = async (lockManager, amount, sender = undefined) => {
```
sohkai, line 36: Hmm ok, so wondering, if these are really meant as optional arguments, perhaps we can use a trailing options argument to mark them as such?

sohkai (approved):
🙌 only small nitpicky comments!

I'm not in love with the diamond inheritance resulting from `Controlled`, `ControlledRecoverable`, and `ControlledRelayer` (didn't realize this would be a problem from #25), but I also don't see a way around it for now. Perhaps we can remove it later if we bundle the role authentication directly into `Controlled`.

docs/4-entry-points/4-guardians-registry.md
```diff
@@ -20,12 +20,41 @@ This module is in the one handling all the staking/unstaking logic for the guard
     - Save the guardian token address
     - Save the total active balance limit
 
-### 4.4.2. Activate
+### 4.4.2. Stake
 
-- **Actor:** Guardian of the Protocol
+- **Actor:** Guardian or an external entity incentivized in a guardian of the Protocol
```
sohkai, line 25: Perhaps we can replace "an external entity incentivized in a guardian of the Protocol" (which doesn't make much sense to me), to just "whitelisted relayer" for now?

PR #27, opened 2020-10-30
Voting: Implement delegate voting with signatures

facuspagnuolo:


sohkai (approved):
LGTM!

The lingering comment I have about this module is with the representatives:

1. Do we need to allow multiple representatives at a time?
2. Could we introduce a `setRepresentative()` that supports signature authorization?

On 1, the only reason I have for disallowing multiple is that it'd allow us to conform to the `delegateBySig()` interface in a wrapper.

facuspagnuolo:
@sohkai I ended addressing both 1 and 2 :)

docs/4-entry-points/5-voting.md
```diff
@@ -14,16 +14,16 @@ In particular, the first version of the protocol uses a commit-reveal mechanism.
...
 
-- **Actor:** Any guardian that could potentially be drafted for an adjudication round
+- **Actor:** Any guardian that could potentially be drafted for an adjudication round or an external entity incentivized in a guardian of the Protocol
 - **Inputs:**
-    - **Representatives:** List of representatives addresses
-    - **Allowed:** Whether each representative is allowed or not
-- **Authentication:** Open
+    - **Voter:** Address of the voter setting their delegate
+    - **Delegate:** Address of the delegate to be set
+- **Authentication:** Open. Implicitly voters on their behalf or a whitelisted relayer.
```
sohkai, line 23:
```suggestion
- **Authentication:** Open. The guardian or a whitelisted relayer.
```

docs/4-entry-points/5-voting.md
```diff
@@ -14,16 +14,16 @@ In particular, the first version of the protocol uses a commit-reveal mechanism.
 - **State transitions:**
     - Save the controller address
 
-### 4.5.2. Set representative
+### 4.5.2. Delegate
 
-- **Actor:** Any guardian that could potentially be drafted for an adjudication round
+- **Actor:** Any guardian that could potentially be drafted for an adjudication round or an external entity incentivized in a guardian of the Protocol
```
sohkai: Similar to #26, I don't understand what "an external entity incentivized in a guardian of the Protocol" means.

Technically, any account is allowed to set a delegate here, so we could probably leave this as just "open".

sohkai (approved):
Just minor nitpicks on documentation, similar to #26's recent comments.

PR #32, opened 2020-11-02
Payments: Support signature methods and multiple fixes

facuspagnuolo:


packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -277,60 +236,56 @@ contract PaymentsBook is ControlledRecoverable, TimeHelpers, IPaymentsBook {
...
+        }
     }
 
     /**
-    * @dev Tell the share corresponding to the governor for a certain period
+    * @dev Tell if the governor can claim the owed share for a certain period
     * @param _periodId Identification number of the period being queried
     * @param _tokens List of token addresses to be queried
-    * @return List of token amounts corresponding to the governor
+    * @return List of status to tell whether the governor can claim the given list of tokens
```
sohkai:
```suggestion
    * @return List of statuses to tell whether the governor can claim the given list of tokens
```

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -383,6 +338,7 @@ contract PaymentsBook is ControlledRecoverable, TimeHelpers, IPaymentsBook {
         if (_token == address(0)) {
             require(msg.value == _amount, ERROR_ETH_DEPOSIT_MISMATCH);
         } else {
+            require(msg.value == 0, ERROR_ETH_DEPOSIT_MISMATCH);
```
sohkai: We can optimize this condition, if we do the `msg.value == 0` check first:

```solidity
if (msg.value > 0) {
  // ETH
  require(_token == address(0));
} else {
  require(_token.transferFrom());
}
```

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -351,8 +306,8 @@ contract PaymentsBook is ControlledRecoverable, TimeHelpers, IPaymentsBook {
     )
         internal
     {
+        _period.claimedGuardians[_guardian][_token] = true;
         if (_amount > 0) {
```
sohkai, line 310: ⚠️ Now that we're reverting if you have already claimed before (`_canGuardianClaim()`), I wonder if we shouldn't also revert if the amount is 0.

It'd be a waste of chain space, but especially for the governor, but it's awkward that someone could keep calling to claim shares of random addresses when we know which tokens and amounts are claimable.

If we do this, I would again push for `canGuardianClaim()` and `canGovernorClaim()` to factor in whether or not the guardian/governor has an amount to claim.

~(There's also a nitpick here that, like `_claimGovernorShare()`, maybe we should move the `_getGuardianShare()` to be called here?)~ I didn't notice that `_getGuardianShare()` required a few additional variables not passed down!

sohkai (commented):
LGTM, other than my note on the error behaviour with claiming a share.

Now that we've moved to an array, I also realized that this could be hard to construct a spec-compliant ERC-712 signature for. From what I've seen, the standard is still missing an authoritative way to support arrays although MetaMask does support it with their `signTypedData_v4` API.

(I'm not sure how the first commit in this PR got included, but I think only the commits after are relevant for review?)

packages/evm/contracts/payments/PaymentsBook.sol
facuspagnuolo: I don't see how it would be optimized tbh, you still need to check that `msg.value == _amount`, so you end up having three `require`s anyway. Also, it sounds weird not checking the token address in the second case.

sohkai: Discussed offline, and good catch on the second require with `msg.value == _amount`.

The thinking is that putting more requires in the ETH branch is better, because:

1. Receiving tokens will likely be more popular
2. The ETH branch is cheaper, so we "smooth out" the cost a bit between the branches.

packages/evm/contracts/payments/PaymentsBook.sol
```diff
@@ -520,19 +475,19 @@ contract PaymentsBook is ControlledRecoverable, TimeHelpers, IPaymentsBook {
...
+    * @dev Tell if a guardian can claim the owed share for a certain period
     * @param _period Period being queried
     * @param _guardian Address of the guardian being queried
     * @param _token Address of the token to be queried
-    * @return True if the guardian has already claimed their share
+    * @return True if the guardian can claim their share
     */
-    function _hasGuardianClaimedShare(Period storage _period, address _guardian, address _token) internal view returns (bool) {
-        return _period.claimedGuardians[_guardian][_token];
+    function _canGuardianClaim(Period storage _period, address _guardian, address _token) internal view returns (bool) {
```
sohkai: Ah yes, were we going to re-flip this and the governor one back to `_hasGuardianClaimedShare()`?

sohkai (approved):

packages/evm/contracts/payments/PaymentsBook.sol
sohkai: After a protracted discussion on the pros and cons of this (unfortunately adding this check above was not so easy because we may need to obtain the randomness for fetching a guardian's balance in any given period), we ultimately decided to re-flip the getter so that it reports whether or not a guardian has already claimed a token.

UIs and contracts will be able to use this getter to know whether a claim request has already been processed, and not request it again. It should be safe, as claim requests resulting in 0 tokens being transferred are still allowed.

PR #33, opened 2020-11-02
Treasury: Implement signature methods

facuspagnuolo:


packages/evm/contracts/treasury/ProtocolTreasury.sol
```diff
@@ -49,26 +50,27 @@ contract ProtocolTreasury is ControlledRecoverable, ITreasury {
...
     */
-    function withdraw(IERC20 _token, address _to, uint256 _amount) external {
-        _withdraw(_token, msg.sender, _to, _amount);
+    function withdraw(IERC20 _token, address _from, address _to, uint256 _amount) external authenticate(_from) {
+        _withdraw(_token, _from, _to, _amount);
     }
 
     /**
-    * @notice Withdraw all the tokens from `_to` to themself
+    * @notice Withdraw all the tokens from `_from` to themself
```
sohkai:
```suggestion
    * @notice Withdraw all the tokens from `_from` to themselves
```

packages/evm/contracts/treasury/ProtocolTreasury.sol
```diff
@@ -49,26 +50,27 @@ contract ProtocolTreasury is ControlledRecoverable, ITreasury {
...
     * @param _token ERC20 token to be withdrawn
-    * @param _to Address of the recipient that will receive their tokens
+    * @param _from Address withdrawing the tokens from and where the tokens will be transferred
     */
-    function withdrawAll(IERC20 _token, address _to) external {
-        IConfig config = _protocolConfig();
-        require(config.areWithdrawalsAllowedFor(_to), ERROR_WITHDRAWS_DISALLOWED);
+    function withdrawAll(IERC20 _token, address _from) external {
+        bool canWithdrawAll = _isSignatureValid(_from) || _protocolConfig().areWithdrawalsAllowedFor(_from);
```
sohkai: I'm not sure I fully understand the difference between `withdraw()` and `withdrawAll()`; why does this one check `_protocolConfig().areWithdrawalsAllowedFor(_from)`?

Furthermore, I'm wondering why this is an `or` and not an `and`. Wouldn't you want to check that the originating account has allowed this whether it's the direct caller or not?

-------

It's a bit more gas, but I'm also wondering if it'd be worth still using the `authenticate()` modifier here.

In this case, I think the signature being wrong should present a different error.

packages/evm/test/treasury/protocol-treasury.js
```diff
@@ -162,194 +164,224 @@ contract('ProtocolTreasury', ([_, disputeManager, holder, someone]) => {
...
+
+            context('when the treasury contract does not have enough tokens', () => {
+              it('reverts', async () => {
+                await assertRevert(withdraw(DAI, recipient, amount, sender, authorize), TREASURY_ERRORS.WITHDRAW_FAILED)
+              })
             })
           })
 
-          context('when the treasury contract does not have enough tokens', () => {
+          context('when the given amount is grater than the balance of the account', () => {
```
sohkai:
```suggestion
          context('when the given amount is greater than the balance of the account', () => {
```

😄

packages/evm/test/treasury/protocol-treasury.js
```diff
@@ -162,194 +164,224 @@ contract('ProtocolTreasury', ([_, disputeManager, holder, someone]) => {
...
             it('reverts', async () => {
-              await assertRevert(treasury.withdraw(DAI.address, recipient, amount, { from }), TREASURY_ERRORS.WITHDRAW_FAILED)
+              await assertRevert(withdraw(DAI, recipient, amount, sender, authorize), TREASURY_ERRORS.WITHDRAW_INVALID_AMOUNT)
             })
           })
         })
 
-        context('when the given amount is grater than the balance of the account', () => {
-          const amount = bigExp(201, 18)
+        context('when the given recipient is the zero address', () => {
```
sohkai, line 286: I found this test kind of weird, do we have any explicit checks on the zero address? If so, should we also check that tokens are not being sent back to the treasury contract?

packages/evm/test/treasury/protocol-treasury.js
```diff
@@ -162,194 +164,224 @@ contract('ProtocolTreasury', ([_, disputeManager, holder, someone]) => {
...
-      beforeEach('accept automatic withdraws', async () => {
-        await controller.setAutomaticWithdrawals(true, { from: recipient })
-      })
+    const withdrawAll = async (token, sender, authorize = false) => {
+      let calldata = treasury.contract.methods.withdrawAll(token.address, holder).encodeABI()
+      if (authorize) calldata = await encodeAuthorization(treasury, holder, getPKForAccount(holder), calldata, sender)
+      return treasury.sendTransaction({ from: sender, data: calldata })
+    }
 
+    const itHandlesWithdrawsProperly = (sender, authorize = false) => {
```
sohkai:
```suggestion
    const itHandlesWithdrawAllsProperly = (sender, authorize = false) => {
```

sohkai (commented):
Main note is just with `withdrawAll()`. I might be missing some context on how it's meant to be used.

packages/evm/contracts/treasury/ProtocolTreasury.sol
```diff
@@ -55,17 +55,17 @@ contract ProtocolTreasury is ITreasury, ControlledRecoverable, SignaturesValidat
...
         _withdraw(_token, _from, _to, _amount);
     }
 
     /**
-    * @notice Withdraw all the tokens from `_from` to themself
+    * @notice Withdraw all the tokens from `_from` to themselves
     * @param _token ERC20 token to be withdrawn
     * @param _from Address withdrawing the tokens from and where the tokens will be transferred
     */
     function withdrawAll(IERC20 _token, address _from) external {
```
sohkai: From our discussion yesterday, perhaps we can remove this function altogether? Since we can re-introduce it at a later time (and perhaps, the config for withdrawals can also be removed since this sort of functionality would be pushed outside).

sohkai (approved):
Aside from the bit on removing `withdrawAll()`, LGTM!

PR #34, opened 2020-11-02
Protocol: Entry point enhancements

facuspagnuolo:


packages/evm/contracts/core/modules/Controller.sol
```diff
@@ -81,6 +83,14 @@ contract Controller is IsContract, ModuleIds, ProtocolClock, ProtocolConfig {
         _;
     }
 
+    /**
+    * @dev Ensure the given dispute manager is active
+    */
+    modifier onlyActiveDisputeManagers(IDisputeManager _disputeManager) {
```
sohkai: Nitpick: perhaps we can just leave this input as an address, since we don't need the interface at all? We don't really gain anything from the typing here, in terms of safety.

packages/evm/contracts/disputes/DisputeManager.sol
```diff
@@ -201,14 +183,29 @@ contract DisputeManager is ControlledRecoverable, ICRVotingOwner, IDisputeManage
...
+    * @param _disputeId Identification number of the dispute receiving new evidence
+    * @param _submitter Address of the account submitting the evidence
+    * @param _evidence Data submitted for the evidence of the dispute
+    */
+    function submitEvidence(IArbitrable _subject, uint256 _disputeId, address _submitter, bytes calldata _evidence)
+        external
+        onlyController
+    {
+        (Dispute storage dispute, ) = _getDisputeAndRound(_disputeId, 0);
+        require(dispute.subject == _subject, ERROR_SENDER_NOT_DISPUTE_SUBJECT);
```
sohkai: Nitpick: perhaps `SUBJECT_NOT_DISPUTE_SUBJECT` is more correct, since the sender is the controller

packages/evm/contracts/disputes/DisputeManager.sol
```diff
@@ -1126,34 +1104,102 @@ contract DisputeManager is ControlledRecoverable, ICRVotingOwner, IDisputeManage
...
 
     /**
-    * @dev Internal function to check if a certain dispute exists, it reverts if it doesn't
-    * @param _disputeId Identification number of the dispute to be checked
+    * @dev Internal function to check if a certain dispute round exists, it reverts if it doesn't
+    * @param _dispute Dispute to be checked
+    * @param _roundId Identification number of the dispute round to be checked
     */
-    function _checkDisputeExists(uint256 _disputeId) internal view {
+    function _checkRoundExists(Dispute storage _dispute, uint256 _roundId) internal view {
```
sohkai, line 1110: Maybe this could be inlined into a `_getDisputeRound()`? All the usages seem to just be interested in the round :).

packages/evm/contracts/disputes/DisputeManager.sol
```diff
@@ -1126,34 +1104,102 @@ contract DisputeManager is ControlledRecoverable, ICRVotingOwner, IDisputeManage
...
+        round = dispute.rounds[_roundId];
+        config = _getDisputeConfig(dispute);
+    }
+
+    /**
+    * @dev Internal function to get the Protocol config used for a dispute
+    * @param _dispute Dispute querying the Protocol config of
+    * @return Protocol config used for the given dispute
+    */
+    function _getDisputeConfig(Dispute storage _dispute) internal view returns (Config memory) {
```
sohkai: Nitpick: maybe it's worth pushing this up to be after `_getDispute()` and `_getDisputeRound()` so all the "singular" versions live together :).

sohkai (commented):
Only nitpicks, this looks really good 🙌 🙌 🙌.

That being said, while I like the changes, I'm wondering how much better you feel about the new internal functions that were added. Do you think they help a lot with readability and safety?

Also wondering if it'd be worth trying to bump the optimizer again, since the removal of a modifier should give us a bit more space now.

facuspagnuolo:
> That being said, while I like the changes, I'm wondering how much better you feel about the new internal functions that were added. Do you think they help a lot with readability and safety?

The main reason was not because of safety or readability but to avoid duplicated code for the compiler, hence reducing the bytecode size

packages/evm/contracts/core/modules/Controller.sol
facuspagnuolo: But we will be moving the cast outside of it, I guess I prefer the abstraction

sohkai (approved):
LGTM ❤

PR #37, opened 2020-11-04
evm(registry): migrate interface to be "ERC20 lite"

sohkai:
Updates various interfaces that were inspired by ERC-900 to now instead be following ERC-20.

Also includes some of ERC-20's metadata fields, like name and symbol, for tooling to pick up on.


packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -108,10 +113,10 @@ contract GuardiansRegistry is IGuardiansRegistry, ControlledRecoverable, Control
     }
 
     // Maximum amount of total active balance that can be held in the registry
-    uint256 internal totalActiveBalanceLimit;
+    uint256 public totalActiveBalanceLimit;
```
sohkai, line 114: Didn't think it really made sense to expose a different getter for this

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -498,50 +495,68 @@ contract GuardiansRegistry is IGuardiansRegistry, ControlledRecoverable, Control
...
-    * @param _guardian Address of the guardian querying the balance information of
+    * @dev Tell the total balance of tokens held by a guardian for a given term id
+    *      This includes the active balance, the available balances, and the pending balance for deactivation.
+    *      Note that we don't have to include the locked balances since these represent the amount of active tokens
+    *      that are locked for drafts, i.e. these are already included in the active balance of the guardian.
+    * @param _guardian Address of the guardian querying the balance of
+    * @param _termId Term ID to query on
+    * @return Total amount of tokens of a guardian
+    */
+    function balanceOfAt(address _guardian, uint64 _termId) external view returns (uint256) {
```
sohkai: ⚠️ I wasn't quite sure about:

1. Where to best test this
2. Whether the logic here is correct. It mimics how `_balanceOf()` works.

packages/evm/.soliumrc.json
```diff
@@ -16,7 +16,7 @@
     "lbrace": "error",
     "mixedcase": "error",
     "camelcase": "error",
-    "uppercase": "error",
+    "uppercase": "off",
```
sohkai, line 19: Fixes the linting issue with `string public constant name`. I don't think this rule is that useful anyway :)

packages/evm/contracts/registry/GuardiansRegistry.sol
facuspagnuolo: 1. I'd say in `guardians-registry-activation` since you will need to query the active balance for different terms
2. I'd it mimics `_balanceOf` but for a given `_termId` right?

sohkai: > I'd it mimics _balanceOf but for a given _termId right?

Yes, that was the intention (since we had the `at()` versions for the detailed and other balance getters).

sohkai: Removed `balanceOfAt()` in a8f7fa7, as we clarified that the only checkpointed balance we maintain is the guardians' active balance.

packages/evm/contracts/registry/GuardiansRegistry.sol
```diff
@@ -106,10 +111,10 @@ contract GuardiansRegistry is IGuardiansRegistry, ControlledRecoverable, Control
     }
 
     // Maximum amount of total active balance that can be held in the registry
-    uint256 internal totalActiveBalanceLimit;
+    uint256 public totalActiveBalanceLimit;
 
     // Guardian ERC20 token
-    IERC20 internal guardiansToken;
```
facuspagnuolo, line 112: WDYT about renaming it to `token` now?

facuspagnuolo (commented):
LGTM!

PR #38, opened 2020-11-05
Controller: Implement ACL

facuspagnuolo:


packages/evm/test/registry/guardians-registry-locking.js
```diff
@@ -268,11 +256,7 @@ contract('GuardiansRegistry', ([_, guardian, someone, governor]) => {
               })
             })
 
-            context('when sender was not a whitelisted relayer', () => {
-              before('disallow relayer', async () => {
```
sohkai, line 272: Was thinking that maybe it'd be helpful to have some sort of permission-setting helper that can just do nothing when you give it `false` and the address doesn't hold the role.

I think it can be helpful in some of these nested scenarios to re-emphasize the test's state, what do you think?

packages/evm/contracts/core/modules/ACL.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    // List of all roles assigned to different addresses
+    mapping (bytes32 => mapping (address => bool)) public roles;
+
+    event Granted(bytes32 indexed id, address indexed who);
+    event Revoked(bytes32 indexed id, address indexed who);
+    event Frozen(bytes32 indexed id);
+
+    modifier notFrozen(bytes32 _id) {
+        require(!isRoleFrozen(_id), ERROR_ROLE_ALREADY_FROZEN);
```
sohkai: Curious if you like this modifier more than just inlining it.

If we do, let's also make the change in Govern's ACL so it's checked in a similar manner.

packages/evm/contracts/core/modules/ACL.sol
```diff
@@ -0,0 +1,96 @@
...
+        require(_op.length == _id.length && _op.length == _who.length, ERROR_INVALID_BULK_INPUT);
+
+        for (uint256 i = 0; i < _op.length; i++) {
+            BulkOp op = _op[i];
+            if (op == BulkOp.Grant) {
+                _grant(_id[i], _who[i]);
+            } else if (op == BulkOp.Revoke) {
+                _revoke(_id[i], _who[i]);
+            } else if (op == BulkOp.Freeze) {
+                _freeze(_id[i]);
```
sohkai, line 92: I guess in this case, you would put in anything you wanted for the `_who` element (most likely the zero address to save calldata gas)?

packages/evm/contracts/core/modules/ACL.sol
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+    * @dev Internal function to grant a role to a given address
+    * @param _id ID of the role to be granted
+    * @param _who Address to grant the role to
+    */
+    function _grant(bytes32 _id, address _who) internal notFrozen(_id) {
+        require(_who != FREEZE_FLAG, ERROR_BAD_FREEZE);
+        require(!hasRole(_who, _id), ERROR_ROLE_GRANTED);
```
sohkai: I was thinking about this for the Govern ACL too, what do you think about not using a require, but skipping the state change and event emission if the user was already granted the role?

sohkai (commented):
Couple questions, but overall this looks great!

packages/evm/contracts/core/modules/ACL.sol
facuspagnuolo: Yup, exactly

packages/evm/test/voting/crvoting-commit.js
```diff
@@ -298,6 +302,10 @@ contract('CRVoting', ([_, voter, someone, delegate, governor]) => {
           })
 
           context('when the sender does not have permission', () => {
+            beforeEach('revoke role', async () => {
+              await controller.grant(roleId(delegate, 'commit'), sender, { from: governor })
```
sohkai: Should this be `revoke()` instead? Although it's a bit weird, shouldn't this make the test fail?

sohkai (approved):

packages/evm/test/voting/crvoting-commit.js
facuspagnuolo: It failed on my computer, this is weird

PR #41, opened 2020-11-05
Registry: remove detailedBalanceOfAt() as only active balance is checkpointed

sohkai:
As discussed in #37, only the active balance is checkpointed so all the other balances should be obtained from `detailedBalanceOf()` instead.

### DIFF
--- a/docs/6-external-interface/4-guardians-registry.md
+++ b/docs/6-external-interface/4-guardians-registry.md
@@ -173,7 +173,7 @@ The following functions are state getters provided by the `GuardiansRegistry`:
     - **Locked:** Amount of active tokens that are locked due to ongoing disputes
     - **Pending deactivation:** Amount of active tokens that were requested for deactivation
 
-#### 6.4.2.8. Active balance of at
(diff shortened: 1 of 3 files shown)


docs/6-external-interface/4-guardians-registry.md
```diff
@@ -173,17 +173,6 @@ The following functions are state getters provided by the `GuardiansRegistry`:
     - **Locked:** Amount of active tokens that are locked due to ongoing disputes
     - **Pending deactivation:** Amount of active tokens that were requested for deactivation
 
-#### 6.4.2.7. Detailed balance of at
```
facuspagnuolo, line 176: I know it's a PITA, but should we update the following indexes?

facuspagnuolo (approved):
LGTM!

docs/6-external-interface/4-guardians-registry.md
sohkai: Ah yes totally, forgot about it!

